PR #9, opened 2024-02-29
[AJUN-197] Season config changes

DidacSF:
## Description

Applies changes described [here](https://ajunanetwork.atlassian.net/wiki/spaces/PLATFORM/pages/225181697/Season+Config+changes)

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/benchmarking/src/lib.rs
+++ b/pallets/ajuna-awesome-avatars/benchmarking/src/lib.rs
@@ -28,9 +28,9 @@ use frame_support::{
 use frame_system::{pallet_prelude::BlockNumberFor, RawOrigin};
 use pallet_ajuna_awesome_avatars::{types::*, Config as AvatarsConfig, Pallet as AAvatars, *};
 use pallet_ajuna_nft_transfer::traits::NftHandler;
-use sp_runtime::{
-	traits::{Saturating, StaticLookup, UniqueSaturatedFrom, UniqueSaturatedInto, Zero},
-	BoundedVec,
+use sp_runtime::bounded_vec;
+use sp_runtime::traits::{
+	Saturating, StaticLookup, UniqueSaturatedFrom, UniqueSaturatedInto, Zero,
 };
 use sp_std::vec;
 
@@ -80,11 +80,6 @@ fn create_seasons<T: Config>(n: usize) -> Result<(), &'static str> {
 		Seasons::<T>::insert(
 			(i + 1) as SeasonId,
 			Season {
-				name: [u8::MAX; 100].to_vec().try_into().unwrap(),
-				description: [u8::MAX; 1_000].to_vec().try_into().unwrap(),
-				early_start: BlockNumberFor::<T>::from((i * 10 + 1) as u32),
-				start: BlockNumberFor::<T>::from((i * 10 + 5) as u32),
-				end: BlockNumberFor::<T>::from((i * 10 + 10) as u32),
 				max_tier_forges: u32::MAX,
 				max_variations: 15,
 				max_components: 16,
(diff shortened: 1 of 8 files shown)


darkfriend77 (approved):
okay

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -291,7 +306,13 @@ pub mod pallet {
...
 		TreasuryClaimed { season_id: SeasonId, treasurer: T::AccountId, amount: BalanceOf<T> },
 		/// The season configuration for {season_id} has been updated.
-		UpdatedSeason { season_id: SeasonId, season: SeasonOf<T> },
+		UpdatedSeason {
+			season_id: SeasonId,
+			season: Option<SeasonOf<T>>,
+			meta: Option<SeasonMeta>,
+			schedule: Option<SeasonScheduleOf<T>>,
+			trade_filters: Option<TradeFilters>,
+		},
```
clangenb, line 315: Hmm, this is interesting, from a client perspective, I wonder if it is more convenient to have separate events? In this case, you can only subscribe to the relevant event, and you don't need extra logic to find out what exactly has changed. Even more so, I guess there are fields that we never really expect to change?

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -872,12 +898,39 @@ pub mod pallet {
 		pub fn set_season(
 			origin: OriginFor<T>,
 			season_id: SeasonId,
-			season: SeasonOf<T>,
+			season: Option<SeasonOf<T>>,
+			season_meta: Option<SeasonMeta>,
+			season_schedule: Option<SeasonScheduleOf<T>>,
+			trade_filters: Option<TradeFilters>,
```
clangenb, line 904: Same as above, I think from a separation of concern perspective, doesn't it make more sense to have separate xts? 

Of have a `set_season` that has mandatory fields and a `update_xxx` to change a specific value?

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -872,12 +898,39 @@ pub mod pallet {
...
 		) -> DispatchResult {
 			Self::ensure_organizer(origin)?;
-			let season = Self::ensure_season(&season_id, season)?;
-			Seasons::<T>::insert(season_id, &season);
-			Self::deposit_event(Event::UpdatedSeason { season_id, season });
+
+			if let Some(mut season_update) = season.clone() {
```
clangenb, line 908: With the `ref` keyword you can omit the `clone` as we only need a borrowed value afterward.

pallets/ajuna-awesome-avatars/src/tests.rs
```diff
@@ -385,147 +388,156 @@ mod season {
...
+				// Season 3 end (block 54..63)
+				for n in (season_3_schedule.end + 1)..=(season_3_schedule.end + 10) {
+					run_to_block(n);
+					assert_eq!(
+						CurrentSeasonStatus::<Test>::get(),
+						SeasonStatus {
+							season_id: 4,
+							early: false,
+							active: false,
+							early_ended: false,
```
clangenb, line 530: To me, it seems that the three options are mutually exclusive. This makes me think that an `enum SeaonTimeline { inactive, early, active, earlyEnded }` is a better choice, as it is less error-prone.

pallets/ajuna-awesome-avatars/benchmarking/src/lib.rs
```diff
@@ -461,13 +469,35 @@ benchmarks! {
...
+		};
+		let trade_filters = TradeFilters(bounded_vec![
+			u32::from_le_bytes([0x11, 0x07, 0x00, 0x00]), // CrazyDude pet
+			u32::from_le_bytes([0x12, 0x36, 0x00, 0x00]), // GiantWoodStick armor front pet part
+			u32::from_le_bytes([0x25, 0x07, 0x00, 0xFF]), // Metals of quantity 255
+			u32::from_le_bytes([0x25, 0x02, 0x00, 0x00]), // Electronics of any quantity
+			u32::from_le_bytes([0x30, 0x00, 0x00, 0x00]), // Any Essence
+			u32::from_le_bytes([0x41, 0x00, 0x00, 0xF0]), // ArmorBase of quantity 240
+			u32::from_le_bytes([0x45, 0x00, 0x00, 0x0F]), // WeaponVersion1 of quantity 15
+		]);
```
clangenb, line 495: What I fail to understand yet, why are we using bytes instead of an enum?

PR #14, opened 2024-03-13
[AJUN-3] Pallet Chapters/Tournaments

DidacSF:
## Description

Implementation of Pallet Chapters/Tournaments as described [here](https://ajunanetwork.atlassian.net/wiki/spaces/PLATFORM/pages/209879041/Season+Chapter+Tournaments).

## Type of changes

## Checklist


pallets/ajuna-affiliates/src/lib.rs
```diff
@@ -101,8 +101,6 @@ pub mod pallet {
 
 	#[pallet::error]
 	pub enum Error<T, I = ()> {
-		/// There is no account set as the organizer
-		OrganizerNotSet,
```
DidacSF, line 105: Removed this since I realized its not used anywhere in the pallet

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+			Ok(reward_table_total_dist)
+		}
+
+		fn try_get_current_tournament_id_for(
+			season_id: &T::SeasonId,
+		) -> Result<TournamentId, DispatchError> {
+			let maybe_id = ActiveTournaments::<T, I>::get(season_id);
+			ensure!(maybe_id.is_some(), Error::<T, I>::NoActiveTournamentForSeason);
+			Ok(maybe_id.unwrap())
+		}
```
clangenb, line 230: In practice this will probably not be overlooked, but I prefer not to see any `unwraps` in runtime code if it can be prevented. We can just return the following here:

```rust
ActiveTournaments::<T, I>::get(season_id).ok_or_else(|| Error::<T,I>::NoActiveTournamenForSeason)
```

clangenb (commented):
Nice, this looks very good! I only have some pedantic notifications on my end. I still have to study the specifications a bit more to be sure that the implementation details match the specs, but it seems good so far!

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+			let mut base = T::PalletId::get();
+			// TODO: Improve logic
+			base.0[0] = (*season_id).saturated_into::<u32>() as u8;
+			base.0[1] = *tournament_id as u8;
+			base.into_account_truncating()
+		}
+
+		fn ensure_valid_tournament(
+			config: &TournamentConfigFor<T, I>,
+		) -> Result<u16, DispatchError> {
```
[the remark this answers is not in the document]
DidacSF: Yeah, but I preferred it like this, since some logic can be related to the pallet state in future improvements.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+			let tournament_id = Self::try_get_current_tournament_id_for(season_id)?;
+			let current_block = <frame_system::Pallet<T>>::block_number();
+
+			if let Some(tournament_config) = Tournaments::<T, I>::get(season_id, tournament_id) {
+				if tournament_config.end <= current_block {
+					let treasury_account =
+						Self::tournament_treasury_account_id(season_id, &tournament_id);
+
+					for (category, player_table) in
+						TournamentRankings::<T, I>::iter_prefix((season_id, tournament_id))
```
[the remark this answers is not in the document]
DidacSF: Yeah the amount is limited to a max of 10 entries per category at much

pallets/ajuna-tournament/src/lib.rs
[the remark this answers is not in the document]
clangenb: Ok, valid reasoning. 👍

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+					.unwrap_or_default();
+
+				if account_payout > 0_u32.into() {
+					T::Currency::transfer(
+						treasury_account,
+						account,
+						account_payout,
+						ExistenceRequirement::AllowDeath,
+					)?;
+				}
```
clangenb, line 315: Is it ok, if this fails midst payout? It will interrupt this function, and also the `try_finish_tournament_function`.

pallets/ajuna-tournament/src/traits.rs
```diff
@@ -0,0 +1,139 @@
...
+type TreasuryAccountDecodec<SeasonId> = (PalletId, [u8; 1], SeasonId, [u8; 1], TournamentId);
+
+impl<SeasonId: Encode> Encode for TournamentTreasuryAccount<SeasonId> {
+	fn encode(&self) -> Vec<u8> {
+		// This codec will fit into the indexers rendering design such that we can
+		// see the treasury accounts as "<pallet_id>/season_id/tournament_id".
+		let data: TreasuryAccountEncodec<SeasonId> =
+			(&self.pallet_id, b"/", &self.season_id, &b"/", &self.tournament_id);
+		data.encode()
+	}
```
clangenb, line 34: Hmm, now I am having doubts about using a struct for this. Maybe it is better to just have functions that create the encoded value. It is not good to implement a non-standard parity-scale-codec as all the libraries that autogenerate types, will fail to properly de-/encode it.

I did this to benefit from the auto-implemented `into_account_truncating` if the type satisfies `TypeId`, `Encode`, `Decode`, but I think now that it is actually better to simply write our own `into_account_truncating` replicating the original one, and just derive encode and decode on this type. What do you think?

clangenb (commented):
Thank you for all the readability fixes! They helped me understand the code much better on the intuitive level. This allowed me to ask some more questions. I haven't made it through the tests yet. :)

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+			if let Some(max_reward) = config.max_reward {
+				ensure!(max_reward > 0_u32.into(), Error::<T, I>::InvalidTournamentConfig);
+			}
+
+			if let Some(fee_perc) = config.take_fee_percentage {
+				ensure!(fee_perc <= 100, Error::<T, I>::InvalidTournamentConfig);
+			}
+
+			let reward_table_total_dist =
+				config.reward_table.iter().fold(0_u16, |a, b| a + (*b as u16));
```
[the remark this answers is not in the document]
DidacSF: Yeah that's why I cast the inputs to `u16` instead so that a user cannot overflow it, and since the check is basically `reward_table_total_dist <= 100` a user cannot provide a valid config where this would overflow later on payout checks too.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+
+		fn process_max_reward_payout(
+			payer: &AccountIdFor<T>,
+			tournament_treasury_account: &AccountIdFor<T>,
+			rank_deposit: BalanceOf<T, I>,
+			max_reward: BalanceOf<T, I>,
+		) -> DispatchResult {
+			let tournament_treasury_balance =
+				T::Currency::free_balance(tournament_treasury_account);
+			let tournament_deposit_left = max_reward.saturating_sub(tournament_treasury_balance);
```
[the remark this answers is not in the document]
DidacSF: This is done so that we can check how close we are the maximum reward that the account can hold.

So for example if the result of the subtraction shows us that we have less than the rank_deposit left to fill the account, the rest of it goes to the global tournament account instead.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+		}
+
+		fn try_get_current_tournament_id_for(
+			season_id: &T::SeasonId,
+		) -> Result<TournamentId, DispatchError> {
+			ActiveTournaments::<T, I>::get(season_id)
+				.ok_or_else(|| Error::<T, I>::NoActiveTournamentForSeason.into())
+		}
+
+		fn process_rank_fee_deposit(
```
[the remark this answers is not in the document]
DidacSF: I've renamed the function to `take_fee_from_rank_deposit` hopefully that makes it easier to understand why the logic looks like it does

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+				payer,
+				&global_treasury_account,
+				fee_deposit,
+				ExistenceRequirement::KeepAlive,
+			)?;
+
+			Ok(base_deposit.saturating_sub(fee_deposit))
+		}
+
+		fn process_max_reward_payout(
```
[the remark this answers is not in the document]
DidacSF: I've renamed the function to `try_deposit_fee_in_tournament_treasury` and added some comments to it, does that help understand its use?

pallets/ajuna-tournament/src/lib.rs
DidacSF: I think this needs a proper meeting to discuss what do we want to do about it.

pallets/ajuna-tournament/src/lib.rs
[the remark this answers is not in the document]
clangenb: Ok, then I think this should be added as a comment. It is very hard to understand this as an uninvolved reader. :)

pallets/ajuna-tournament/src/traits.rs
clangenb: I suggest we simply write a function `into_treasury_account_truncating`. I am not even sure if we ever need the "decode" part.

pallets/ajuna-tournament/src/lib.rs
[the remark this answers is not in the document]
clangenb: Thank you, this helps a lot. It also helps me checking that the code does, what it is supposed to do. :)

pallets/ajuna-tournament/src/lib.rs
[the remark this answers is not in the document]
clangenb: wonderful!

pallets/ajuna-tournament/src/traits.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Debug, PartialEq)]
+pub struct TournamentConfig<BlockNumber, Balance> {
+	pub start: BlockNumber,
+	pub end: BlockNumber,
+	pub initial_reward: Option<Balance>,
+	pub max_reward: Option<Balance>,
+	pub take_fee_percentage: Option<u8>,
+	pub reward_table: RewardTable,
+	pub max_players: u32,
+}
```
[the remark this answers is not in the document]
clangenb: Very good comments, they help me a lot!

darkfriend77 (approved):
good, let's test it.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+					ranker.rank_against((entity_id, entity), (other_id, other))
+				}) {
+					match Self::try_update_rank_table(
+						table,
+						&tournament_config,
+						index,
+						entity_id,
+						entity,
+					) {
+						None => Ok(()),
```
clangenb, line 600: This is where I am confused about the semantics as mentioned before.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+			R: EntityRank<EntityId = T::EntityId, Entity = T::RankedEntity>,
+		{
+			let tournament_id = Self::try_get_active_tournament_id_for(season_id)?;
+			let tournament_config = Self::get_active_tournament_config_for(season_id)
+				.ok_or::<Error<T, I>>(Error::<T, I>::TournamentNotFound)?;
+
+			TournamentRankings::<T, I>::mutate(season_id, tournament_id, |table| {
+				if let Err(index) = table.binary_search_by(|(other_id, other)| {
+					ranker.rank_against((entity_id, entity), (other_id, other))
+				}) {
```
clangenb, line 592: Maybe the whole thing is a bit better readable if the show the ok statement before:

```rust
match table.binary_search_by(|(other_id, other)| ranker.rank_against((entity_id, entity), (other_id, other))) {
   Ok(_) => Ok(()),
   Err(index) => {....}
}
```

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+					*state = GoldenDuckState::Enabled(*payout_perc, Some(entity_id.clone()));
+
+					Self::deposit_event(crate::pallet::Event::<T, I>::EntityBecameGoldenDuck {
+						season_id: *season_id,
+						tournament_id,
+						entity_id: entity_id.clone(),
+					});
+				},
+				_ => {},
+			});
```
clangenb, line 647: Reduce duplicate code and make more readable:


```rust
GoldenDuckState::Enabled(payout_perc, maybe_entry_id) => {
			if  maybe_entry_id.map(|e| entity_id < e).unwrap_or(true) =>
		{
			*state = GoldenDuckState::Enabled(*payout_perc, Some(entity_id.clone()));

			Self::deposit_event(crate::pallet::Event::<T, I>::EntityBecameGoldenDuck {
				season_id: *season_id,
				tournament_id,
				entity_id: entity_id.clone(),
			});
		},

```

clangenb: not 100% sure if it is more understandable, now that I read my code XD

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+					{
+						let tournament_config = Tournaments::<T, I>::get(season_id, tournament_id)
+							.ok_or(Error::<T, I>::TournamentNotFound)?;
+						let treasury_account = Self::tournament_treasury_account_id(*season_id);
+
+						let payout_percentage = tournament_config
+							.reward_distribution
+							.into_iter()
+							.nth(index)
+							.unwrap_or_default();
```
clangenb, line 678: I see you are trying to remove the need for copy, but maybe `reward_distribution.get(index).unwrap_or_default()`, followed by a deref op is better? Not sure, just giving ideas here.

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+			entity_id: &T::EntityId,
+		) -> DispatchResult {
+			match ActiveTournaments::<T, I>::get(season_id) {
+				TournamentState::Inactive => Err(Error::<T, I>::NoActiveTournamentForSeason.into()),
+				TournamentState::ActivePeriod(_) =>
+					Err(Error::<T, I>::TournamentNotInClaimPeriod.into()),
+				TournamentState::ClaimPeriod(tournament_id, reward_pot) => {
+					if let Some(index) = TournamentRankings::<T, I>::get(season_id, tournament_id)
+						.iter()
+						.position(|(entry_id, _)| entry_id == entity_id)
```
clangenb, line 668: I prefer to remove the nesting here with:

```rust
let index = TournamentRankings::<T, I>::get(season_id, tournament_id)
						.iter()
						.position(|(entry_id, _)| entry_id == entity_id)
						unwrap_or_else(|| Error::<T, I>::RankingCandidateNotInWinnerTable.into())?;

```

clangenb (commented):
I did quite a detailed review of the lib file, but still haven't had the time to look at the tests XD.

Great job with the refactoring, it looks good overall!

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+			})?;
+
+			Ok(())
+		}
+
+		fn try_get_active_tournament_id_for(
+			season_id: &T::SeasonId,
+		) -> Result<TournamentId, DispatchError> {
+			match ActiveTournaments::<T, I>::get(season_id) {
+				TournamentState::Inactive => Err(Error::<T, I>::NoActiveTournamentForSeason.into()),
```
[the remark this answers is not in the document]
DidacSF: Not really, for the UX its not relevant information, since it uses mostly the events to keep track of the tournament state

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+				);
+
+				Self::deposit_event(Event::<T, I>::TournamentActivePeriodStarted {
+					season_id,
+					tournament_id,
+				});
+
+				T::DbWeight::get().reads_writes(1, 1)
+			} else {
+				T::DbWeight::get().reads(1)
```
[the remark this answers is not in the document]
DidacSF: Yeah, it is a bug if we reach this branch, I'll add a log then

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -0,0 +1,753 @@
...
+				}
+
+				if table.force_insert_keep_left(index, (entity_id.clone(), entity.clone())).is_ok()
+				{
+					Some(true)
+				} else {
+					Some(false)
+				}
+			} else {
+				None
```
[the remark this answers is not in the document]
DidacSF: None basically means that the entity didn't enter the ranking table, it's not really an error since you can try to rank an entity that simply put doesn't make the cut, the only error here is the `Some(false)` because that means you should have inserted the entity in the table but you were not able.

pallets/ajuna-tournament/src/lib.rs
DidacSF: As mentioned in my answer above, None basically means that the entity didn't qualify for ranking, so its not an error, basically we have three possible states for ranking:
* Entity doesn't qualify for ranking -> `None`
* Entity qualifies and table was updated -> `Some(true)`
* Entity qualifies and table failed to update -> `Some(false)`

Of the three only `Some(false)` is an error, and its something that should never happen also, but we have to take into account how the updating of the table works.

pallets/ajuna-tournament/src/lib.rs
DidacSF: I tired to reduce duplication as much as I could with the new commits, please take a look!

darkfriend77 (approved):
go

pallets/ajuna-tournament/src/lib.rs
clangenb: Yeah, the issue is that the data structure makes the code handling difficult here, we mix a setting `Disabled`/`Enabled` here with the state (maybe winner). We can make the code definitely more ergonomic by splitting the state into two separate things to correctly have separation of concerns of config & state. What was your reasoning for putting it into the same storage? Reduce db-reads? I think would only optimize on that level if we run into performance bottlenecks if it makes the code more complicated, but I guess this is opinionated, and I can see where you are coming from.

A good compromise here might be to hide the internal complexity in a setter/getter:


```rust
impl GoldenDuckState {
   pub fn winner() -> Option<EntityId>;

   // updates the winner if enabled and entity_id > entry_id
   pub fn maybe_update_winner()
}

```

pallets/ajuna-tournament/src/lib.rs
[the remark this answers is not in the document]
clangenb: I see, and I think that your comments have greatly helped me to understand what happens. Personally, I think we can make the code even better to understand if we introduce types that carry such meaning directly, I prepared a suggestion in https://github.com/ajuna-network/ajuna-pallets/pull/21

pallets/ajuna-tournament/src/lib.rs
clangenb: If we mix it in one type instead of two than I would suggest to at least separate state and setting within the type aka:

```rust
struct GoldenDuckState {
   pub config: GoldenDuckConfig,
   // contains a potential winner 
   winner: Option<EntityId>,
}

impl GoldenDuckState {
   pub fn new(config: GoldenDuckConfig) -> Self {
      Self { config, winner: None }
   }

   pub fn set_winner(&mut self, winner: EntityId) -> Result<(), ()> {
      match self.config {
          Enabled => {
             self.winner = Some(winner);
             Ok(())
          },
          Disabled => Err(())
     }
   }

   pub fn winner() -> Option<EntityId>
}

```

PR #15, opened 2024-03-15
Implement logic for tournament treasury account id derivation

clangenb:


pallets/ajuna-tournament/src/lib.rs
```diff
@@ -182,14 +182,15 @@ pub mod pallet {
 
 		/// The account ID of the sub-account for a given season_id/tournament_id.
 		pub fn tournament_treasury_account_id(
-			season_id: &T::SeasonId,
-			tournament_id: &TournamentId,
+			season_id: T::SeasonId,
+			tournament_id: TournamentId,
```
clangenb, line 186: If we dereference the values inside, I prefer not having borrowed values as function args.

DidacSF:
I like the idea! Now the problem though is that for testing this doesn't work, that's because the AccountId is only 8 bytes in the testing runtime, so the bytes get truncated at the PalletId, meaning all accounts are the same.

How would you solve this? Maybe `#cfg[test]` flag for a mock implementation in tests? Or is there a better solution?

clangenb:
Hmm, is there a specific reason that the account ID is only 8 bytes? If we want to reflect the actual production runtime behavior, I think we should make the account ID in this case be the same as the in the production runtime. TBH, I think this would be the best solution, although I haven't researched the consequences of such a change.

Otherwise, I think we could make the `TreasuryAccount` an associated type and create an associated type for its constructor method. I prefer if `cfg(test)` is only used to add functionality; changing functionality is dangerous in my opinion, as you no longer test the production behavior.

DidacSF:
Added the change so that `MockAccountId` is more aligned with the expected production type, with this all tests pass, so I'll merge if you agree @clangenb

clangenb:
Yes, sounds perfect!

PR #16, opened 2024-03-17
use crates.io for polkadot/substrate deps

clangenb:
They do it this way in the fellowship repository, and I think crates.io deps are in general simpler to manage. Using crates.io deps in the nodes, will also remove the need to fork the orml repository, as we don't have to deal with the nested cargo workspaces.

[psvm](https://github.com/paritytech/psvm) makes it incredibly easy to manage polkadot-sdk deps, I simply ran `psvm -v "1.7.0" ` to get the dependencies corresponding to polkadot-v1.7.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,18 +144,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ark-bls12-377-ext"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "20c7021f180a0cbea0380eba97c2af3c57074cdaffe0eef7e840e1c9f2841e55"
-dependencies = [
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -2248,7 +2003,7 @@ dependencies = [
  "scale-info",
  "sp-io",
  "sp-runtime",
- "sp-std 14.0.0 (git+https://github.com/paritytech/polkadot-sdk.git?branch=release-polkadot-v1.7.0)",
+ "sp-std",
```
clangenb, line 2006: It also shows that we had duplicate dependencies before.

darkfriend77 (approved):
good

clangenb:
Superseded by #35

PR #21, opened 2024-03-23
[tournaments] add `RankingResult` to simplify the code a bit

clangenb:


pallets/ajuna-tournament/src/lib.rs
```diff
@@ -173,6 +174,7 @@ pub mod pallet {
 			season_id: T::SeasonId,
 			tournament_id: TournamentId,
 			entity_id: T::EntityId,
+			rank: Rank,
```
clangenb, line 177: I think having the rank here doesn't hurt.

PR #22, opened 2024-03-25
[Tournament] Logic adjustments

DidacSF:
## Description

* Added additional methods and checks to ensure that ranking and price claiming can only be performed at the correct tournament periods.
* Added test to verify that duplicated entities cannot be ranked together

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-tournament/src/traits.rs
+++ b/pallets/ajuna-tournament/src/traits.rs
@@ -1,4 +1,5 @@
 use super::{TournamentConfig, TournamentId};
+use frame_support::pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo};
 use sp_runtime::{traits::Member, DispatchError, DispatchResult};
 
 pub trait EntityRank {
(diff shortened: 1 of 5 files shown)


darkfriend77 (approved):
good

pallets/ajuna-tournament/src/traits.rs
```diff
@@ -12,11 +13,21 @@ pub trait EntityRank {
...
 }
 
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Debug, Default, PartialEq)]
+pub enum TournamentPeriod {
+	#[default]
+	Inactive,
+	Active,
+	Claim,
+}
+
```
clangenb, line 23: When, I look at this, why don't we rewrite the `TournamentStruct` like this?

```rust
#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Debug, Default, PartialEq)]
pub struct TournamentState<Balance> {
	pub id: TournamentId,
	pub reward: Balance,
	pub period: TournamentPeriod
}

pub enum TournamentPeriod {
	#[default]
	Inactive,
	Active,
	Claim,
}
```

It seems like a more intuitive to me than proxying the `TournamentState` enum to the `TournamentPeriod`.

clangenb (commented):
Looks good, but I have one low level question.

pallets/ajuna-tournament/src/traits.rs
clangenb: I am usually careful when different enum variants contain different fields, it is always a bit of a hassle to handle them. So I only do that when I can't avoid it.

DidacSF: In the end I decided to just remove the TournamentPeriod on just use the TournamentState one, since adopting the struct solution meant complicating a bit more the logic in another parts of the code.

PR #25, opened 2024-03-27
Implement generic orml benchmarks

clangenb:
The ORML pallets use a different benchmarking paradigm, which doesn't really go well hand in hand with the one defined in substrate. So I have rewritten the orml-benchmarks to be generic, now we can use them the same way as we can use substrate benchmarks in the node.

This PR builds upon #16, so it should be merged first.


clangenb:
Superseded by #45

PR #36, opened 2024-04-09
[AJUN-278] Ranker Improvement

DidacSF:
## Description

* Expands the `AvatarRankingCategory` enum to provide more ranking categories

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/src/types/avatar/tournament.rs
+++ b/pallets/ajuna-awesome-avatars/src/types/avatar/tournament.rs
@@ -28,6 +28,14 @@ where
 	type EntityId = Id;
 	type Entity = Avatar<BlockNumber>;
 
+	fn can_rank(&self, entity: (&Self::EntityId, &Self::Entity)) -> bool {
+		match self.category {
+			AvatarRankingCategory::MinSoulPointsWithForce(ref force) |
+			AvatarRankingCategory::MaxSoulPointsWithForce(ref force) => entity.1.force() == force.as_byte(),
+			_ => true,
+		}
+	}
+
 	fn rank_against(
 		&self,
 		entity: (&Self::EntityId, &Self::Entity),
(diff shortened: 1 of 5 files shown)


pallets/ajuna-awesome-avatars/src/types/avatar/tournament.rs
```diff
@@ -41,6 +46,35 @@ where
...
+				AvatarRankingCategory::DnaDescending => entity.1.dna.cmp(&other.1.dna).reverse(),
+				AvatarRankingCategory::MinSoulPointsWithForce(ref force) =>
+					if entity.1.force() == force.as_byte() && entity.1.force() == other.1.force() {
+						entity.1.souls.cmp(&other.1.souls).reverse()
+					} else {
+						// Returning Ordering::Equal makes that entity not get ranked in the table
+						// in both the case the table is still empty, or when comparing to another
+						// entity already in the ranks
+						Ordering::Equal
+					},
```
clangenb, line 59: Hmm, I see your point. I don't really see another solution here.

Regardless, when I see that we kind of must hack a way in existing data structures to make things work, I do wonder if another approach might be better. Here, we basically say that the avatars are equal in case they have a different force type. Does this make sense? Or is this not supposed to happen in a production setup, and the `Equal` is just a safe fallback value?

Maybe, we can implement order on the `Force` itself?

DidacSF: This will happen in production yes, since when you define the force ranker you will set which force you want to rank, the only other solution I can see is that we implement another method for asking if an entity should be ranked or not instead of jumbling all together in the rank method, maybe that would be better overall?

clangenb: I think having a separate method does make sense. I would still suggest that we invoke that `can_be_ranked` method internally, and change the method signature to return an `Option<Ordering>` if the entity can't be ranked. Even if we call this method only internally, I think that extracting the method gives us the freedom to even to implement it arbitrarily in other contexts, and we can decide under what circumstances and entity can be ranked. However, then we'd have to make the ranker generic over a ranker filter to separate the ranking function from the ranking filter, which might be an overkill after all. XD

I let you decide, but I think it makes sense to:
* Return an `Option<Ordering>` and the outside can decide what happens in the case of `None`
* Extract a `can_be_ranked` function, which also improves the readability of the code.

clangenb: I like the changes, much more clear!

What made you decide to call the `can_rank` function on the outside instead of calling it on the inside and returning an `Option<Ordering>`?

DidacSF: The main motivator was the fact that I needed the ranking function to match the signature expected by rust's `binary_search` method. thats why I decide to split the check and ordering into two separate functions instead of making the ordering return an `Option`

clangenb: Alright, sounds legit. :D

clangenb (approved):
looks good to me, thanks for answering the questions!

PR #38, opened 2024-04-10
[AJUN-289] Freezeing is limited to the stash size of avatars

DidacSF:
## Description

This PR changes the way we managed locking an avatar so that we can lock an unlimited number of avatars, compared to the previous version in which we were limited by the same limitations players have in their avatar storage.

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/src/lib.rs
+++ b/pallets/ajuna-awesome-avatars/src/lib.rs
@@ -491,6 +491,8 @@ pub mod pallet {
 		AvatarInTrade,
 		/// The avatar is currently locked and cannot be used.
 		AvatarLocked,
+		/// The avatar is not currently locked and cannot be unlocked.
+		AvatarNotLocked,
 		/// The avatar is currently unlocked and cannot be locked again.
 		AvatarUnlocked,
 		/// Tried to forge avatars from different seasons.
@@ -1048,7 +1050,7 @@ pub mod pallet {
 		///
 		/// The origin of this call must specify an avatar, owned by the origin, to prevent it from
 		/// forging, trading and transferring it to other players. When successful, the ownership of
-		/// the avatar is transferred from the player to the pallet's technical account.
+		/// the avatar is removed from the player.
 		///
 		/// Locking an avatar allows for new
 		/// ways of interacting with it currently under development.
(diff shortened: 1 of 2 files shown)


pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -1063,12 +1065,7 @@ pub mod pallet {
 			Self::ensure_unlocked(&avatar_id)?;
 			ensure!(Preparation::<T>::contains_key(avatar_id), Error::<T>::NotPrepared);
 
-			Self::do_transfer_avatar(
-				&player,
-				&Self::technical_account_id(),
-				&avatar.season_id,
-				&avatar_id,
-			)?;
+			Self::try_remove_avatar_ownership_from(&player, &avatar.season_id, &avatar_id)?;
```
darkfriend77, line 1068: It feels a little bit strange to put them into the limbus with no owner.

darkfriend77 (approved):

pallets/ajuna-awesome-avatars/src/lib.rs
DidacSF: Well, its is the simplest solution without adding new additional storage, and allows unlimited avatar locking just like we wanted

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -2097,6 +2089,44 @@ pub mod pallet {
...
+					avatars.len() <=
+						PlayerSeasonConfigs::<T>::get(player, season_id).storage_tier as usize,
+					Error::<T>::MaxOwnershipReached
+				);
+				Ok::<_, DispatchError>(())
+			})?;
+
+			Avatars::<T>::try_mutate(avatar_id, |maybe_avatar| -> DispatchResult {
+				let (from_owner, _) = maybe_avatar.as_mut().ok_or(Error::<T>::UnknownAvatar)?;
+				*from_owner = player.clone();
```
clangenb, line 2125: We don't want to check the current here? Shouldn't this only be used if the current owner is the technical account?

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -2097,6 +2089,44 @@ pub mod pallet {
 			});
 		}
 
+		fn try_remove_avatar_ownership_from(
```
clangenb, line 2092: Maybe we want to call this `try_mover_avatar_ownership_to_technical_account`?

pallets/ajuna-awesome-avatars/src/lib.rs
DidacSF: We already check in `unlock_avatar` if the avatar is owned by the technical account, so the check here is not necessary.

pallets/ajuna-awesome-avatars/src/lib.rs
DidacSF: Hmm well both names are fine, since the technical account doesn't own the avatar per se, it mostly holds it, since it cannot do anything with it anyways.

PR #44, opened 2024-04-21
Add multiblock migration for the AAA pallet

### LINKED ISSUE #42: Implement `SteppedMigration` for AAA
The pallet-migrations needs to be added to the runtime and we need to change the migrations here to implement the stepped migrations trait, example here: https://github.com/paritytech/polkadot-sdk/blob/305d311d5c732fcc4629f3295768f1ed44ef434c/substrate/frame/examples/multi-block-migrations/src/migrations/v1/mod.rs

clangenb:
Implement multi-block migrations for AAA according to #42. Closes #42.


DidacSF (approved):

pallets/ajuna-awesome-avatars/src/migration/v6/mod.rs
```diff
@@ -0,0 +1,784 @@
...
+			// problematic if it is the first migration that ran in this block. But there is nothing
+			// that we can do about it here.
+			if meter.remaining().any_lt(required) {
+				return Err(SteppedMigrationError::InsufficientWeight { required });
+			}
+
+			let mut migration_count = 0u32;
+			// We loop here to do as much progress as possible per step.
+			loop {
+				if meter.try_consume(required).is_err() {
```
ggwpez, line 490: This loop was probably a bit annoying to write, or?  
I want to change this `step` function to already do the looping, that you can focus on the migration step itself.  

You think this would make it better or worse @clangenb ?

clangenb: I have actually been thinking about the same thing, I think it would improve things, and reduce the total lines of code written. :)

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -259,6 +259,12 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// This is only an intermediate storage that is being used during the multiblock runtime
+	/// migration of v5 to v6. It should be removed afterward.
+	#[pallet::storage]
+	pub type TradeStatsMap<T: Config> =
+		StorageDoubleMap<_, Identity, SeasonId, Identity, T::AccountId, (Stat, Stat), OptionQuery>;
```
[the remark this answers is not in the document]
clangenb: Ah, I see. I didn't investigate what the macro does exactly, so I wasn't aware of all the options there. Thanks a lot, I have learned something new today. 🚀

PR #45, opened 2024-04-25
Implement generic orml benchmarks

clangenb:
The ORML pallets use a different benchmarking paradigm, which doesn't really go well hand in hand with the one defined in substrate. So I have rewritten the orml-benchmarks to be generic, now we can use them the same way as we can use substrate benchmarks in the node.

Tested that the benchmarks work in
* https://github.com/ajuna-network/bajun-parachain/pull/24
* https://github.com/ajuna-network/ajuna-parachain/pull/15

### DIFF
--- a/orml-pallets-benchmarking/src/vesting.rs
+++ b/orml-pallets-benchmarking/src/vesting.rs
@@ -29,7 +29,7 @@ use orml_vesting::{Call, VestingSchedule};
 
 pub struct Pallet<T: Config>(orml_vesting::Pallet<T>);
 
-pub trait Config: orml_vesting::Config + pallet_balances::Config {}
+pub trait Config: orml_vesting::Config {}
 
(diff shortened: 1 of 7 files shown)


orml-pallets-benchmarking/src/vesting.rs
```diff
@@ -0,0 +1,121 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
```
DidacSF, line 17: Should this notice not be from Ajuna?

clangenb: Wow, awesome catch, will update that!

PR #52, opened 2024-05-27
[AJUN-422] Fix: set locks to true in storage migration for Season 1/2

DidacSF:
## Description

Adjusts `PlayerSeasonConfig` migration so that locks for seasons 1 and 2 are not disabled

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/src/migration/v6/mod.rs
+++ b/pallets/ajuna-awesome-avatars/src/migration/v6/mod.rs
@@ -267,14 +267,11 @@ mod v5 {
 	}
 
 	impl<BlockNumber> PlayerSeasonConfigV5<BlockNumber> {
-		pub fn migrate_to_v6(
-			self,
-			default_locks: Option<Locks>,
-		) -> PlayerSeasonConfig<BlockNumber> {
+		pub fn migrate_to_v6(self) -> PlayerSeasonConfig<BlockNumber> {
 			PlayerSeasonConfig {
 				storage_tier: self.storage_tier,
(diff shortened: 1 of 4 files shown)


pallets/ajuna-awesome-avatars/src/migration/v6/mod.rs
```diff
@@ -521,10 +524,16 @@ pub mod mbm {
 						(old_config.stats.trade.bought, old_config.stats.trade.sold),
 					);
 
+					let default_locks = match season_id {
+						1 | 2 =>
+							Some(Locks { avatar_transfer: true, set_price: true, affiliate: false }),
+						_ => None,
+					};
```
clangenb, line 531: Ah do we have more seasons already, or is this just a safety check?

clangenb (approved):
Looks good, but I prefer having a PR name that actually explains what has been fixed, like: `Fix: set locks to true in storage migration for Season 1/2`

pallets/ajuna-awesome-avatars/src/migration/v6/mod.rs
DidacSF: We have 2 previous seasons this are the two in which the changes need to be made

clangenb: Alright, yeah, this is why I thought that we can just always set it to true, instead of checking the season it. But all good. 👍

darkfriend77: That looks very complex for something that only needs to be done during the migration.

PR #60, opened 2024-07-16
[AJUN-566] Adjust V3 matching logic + additional documentation

DidacSF:
* Adjusted V3 algorithm logic

## Description

Please describe your pull request and link relevant issues.

## Type of changes

- [x] `docs`: Changes to documentation only

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
+++ b/pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
@@ -1064,6 +1064,10 @@ where
 		(avatar.dna[21..32]).copy_from_slice(&value);
 	}
 
+	/// Checks if the `array_1` can be upgraded using `array_2`
+	///
+	/// The method returns an optional list of the indexes that matched and should be used for
+	/// upgrading `array_1`. If there was no match `None` is returned.
 	pub fn is_progress_match(
 		array_1: [u8; 11],
 		array_2: [u8; 11],
@@ -1076,6 +1080,18 @@ where
 		(match_count > 0 && (((match_count * 2) + mirror_count) >= 6)).then_some(matches)
 	}
 
+	/// Matches two progress arrays, using `array_2` to upgrade `array_1`.
+	///
+	/// This method return the indexes that matched and the indexes that are considered mirrors.
+	///
+	/// * Matches are indexes that can be increased to the next rarity level.
+	/// * Mirrors are indexes that have been maxed and are of the same variation between both
+	///   arrays.
+	///
+	/// `rarity_level` is used to determine the maximum rarity for which a direct byte match is not
+	/// necessary for a matching to occur, the higher the value, the easier it is to generate
+	/// matches between two arrays, while at the minimum level (which is 0) only direct byte matches
+	/// or special variations will generate a match.
 	pub fn match_progress(
 		array_1: [u8; 11],
 		array_2: [u8; 11],
@@ -1114,6 +1130,13 @@ where
 		(mirrors, matches)
 	}
 
+	/// Matches two bytes, matching is in short checking if the two bytes are "neighbours"
+	///
+	/// Neighbouring bytes are either those with a difference in value of either 1 or
+	/// `PROGRESS_VARIATIONS`.
+	///
+	/// If there is a match, that means that the index these bytes are from may be selected for
+	/// upgrading the leader's progress_array.
 	pub fn match_progress_byte(byte_1: u8, byte_2: u8) -> bool {
 		let diff = if byte_1 >= byte_2 { byte_1 - byte_2 } else { byte_2 - byte_1 };
 		diff == 1 || diff == (PROGRESS_VARIATIONS - 1)
--- a/pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
+++ b/pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
@@ -1064,55 +1064,56 @@ where
 		(avatar.dna[21..32]).copy_from_slice(&value);
 	}
 
-	/// Checks if the `array_1` can be upgraded using `array_2`
+	/// Checks if the `leader_progress` can be upgraded using `sacrifice_progress`
 	///
 	/// The method returns an optional list of the indexes that matched and should be used for
-	/// upgrading `array_1`. If there was no match `None` is returned.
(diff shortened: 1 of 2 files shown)


pallets/ajuna-awesome-avatars/src/types/avatar/versions/v3.rs
```diff
@@ -208,16 +208,20 @@ impl<T: Config> ForgerV3<T> {
 		}
 
 		let (matching_indexes, match_count, mirror_count) =
-			target.dna.clone().into_iter().zip(other.dna.clone()).enumerate().fold(
+			array_1.into_iter().zip(array_2).enumerate().fold(
```
darkfriend77, line 211: why do we change name here?

darkfriend77 (commented):
I don't get the name change? It's sounds much less verbose now

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v3.rs
```diff
@@ -227,9 +231,13 @@ impl<T: Config> ForgerV3<T> {
 					if have_same_rarity &&
 						!is_maxed && (rarity_1 < max_tier || variation_2 == 0x0B || byte_match)
 					{
+						// if same rarity and on lowest rarity check if gene is a matching
+						// candidate, imp. matching is not equal mirror
```
clangenb: what does imp. 'mean'?

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v3.rs
DidacSF: Because we want to compare only the progress_arrays part of the dna, not the whole dna

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
```diff
@@ -1064,6 +1064,10 @@ where
 		(avatar.dna[21..32]).copy_from_slice(&value);
 	}
 
+	/// Checks if the `array_1` can be upgraded using `array_2`
```
clangenb: Maybe we can find some good naming here to make it clear. I know it is tedious, but I think it is worth it. How about something like:

* `upgradable` - `input`
* I have heard the term sacrifice, I think it could fit here too: `upgradable` - `sacrifice`?
* We talk about a progress match, so maybe we could also use `progressible`?

Any other ideas?

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
```diff
@@ -1076,6 +1080,18 @@ where
 		(match_count > 0 && (((match_count * 2) + mirror_count) >= 6)).then_some(matches)
 	}
 
+	/// Matches two progress arrays, using `array_2` to upgrade `array_1`.
+	///
+	/// This method return the indexes that matched and the indexes that are considered mirrors.
+	///
+	/// * Matches are indexes that can be increased to the next rarity level.
+	/// * Mirrors are indexes that have been maxed and are of the same variation between both
+	///   arrays.
```
clangenb: I don't think the mirrors have to be maxed, are they? I thought that mirrors are when the rarity level of the `progressive`-byte is at higher or equal to the `sacrifices`'s rarity level?

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
```diff
@@ -1114,6 +1130,13 @@ where
 		(mirrors, matches)
 	}
 
+	/// Matches two bytes, matching is in short checking if the two bytes are "neighbours"
+	///
+	/// Neighbouring bytes are either those with a difference in value of either 1 or
+	/// `PROGRESS_VARIATIONS`.
```
clangenb: Maybe we can add that this basically is a module operation.

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
clangenb: I remember now that you use the term `leader` and `sacrifice`? Maybe we can use these terms here too? I know that the functions are more generic, but if they are exclusively used in this context, I think it makes sense to use the domain names.

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
```diff
@@ -1076,6 +1080,18 @@ where
...
+	/// This method return the indexes that matched and the indexes that are considered mirrors.
+	///
+	/// * Matches are indexes that can be increased to the next rarity level.
+	/// * Mirrors are indexes that have been maxed and are of the same variation between both
+	///   arrays.
+	///
+	/// `rarity_level` is used to determine the maximum rarity for which a direct byte match is not
+	/// necessary for a matching to occur, the higher the value, the easier it is to generate
+	/// matches between two arrays, while at the minimum level (which is 0) only direct byte matches
+	/// or special variations will generate a match.
```
clangenb: My brain's sentence cache run out: XD

```suggestion
	/// `rarity_level` is used to determine the maximum rarity for which a direct byte match is not
	/// necessary for a matching to occur. The higher the value, the easier it is to generate
	/// matches between two arrays. At the minimum level (which is 0) only direct byte matches
	/// or special variations will generate a match.
```

clangenb (commented):
Some minor comments, but I like the doc in general! It helps me already a lot! Thanks

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v3.rs
DidacSF: its just a typo, nothing important

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
DidacSF: Yeah, makes sense to adjust it like that

pallets/ajuna-awesome-avatars/src/types/avatar/versions/v2/avatar_utils.rs
DidacSF: You are right, I'll adjust the comments

clangenb (approved):
Nice, thanks a lot for addressing my nitpicks!

PR #62, opened 2024-07-23
Affiliatee chain scripts 

DidacSF:
## Description

This PR contains the scripts and data from the run of `22 July 2024`, which can be used to fix the bug in the `Affiliatee` storage previous to runtime `703`

## Type of changes

## Checklist


affiliate-fix-scripts/generate-event-accounts.py
```diff
@@ -0,0 +1,71 @@
...
+
+        affiliator = ""
+        affiliatee = ""
+
+        if event_params[0]['name'] == 'to':
+            affiliator = event_params[0]['value']
+            affiliatee = event_params[1]['value']
+        else:
+            affiliator = event_params[1]['value']
+            affiliatee = event_params[0]['value']
```
clangenb, line 53: Can you explain this to me, shouldn't they all be identical?

DidacSF: Yeah, I was just paranoid that the response may not guarantee order, that's why I made this check

clangenb:
Thanks a lot, we have executed the fixes after some running a diff to the actual onchain data. Then we only batched the missing affiliate chains into one single batch transaction and executed it here: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fbajun.api.onfinality.io%2Fpublic-ws#/explorer/query/0x23e2c2b1b544d1a6af32d24bfaf5c8de157f96ef7abd3af80db68d869f1fe718

A subsequent diff-check of the onchain data to the .csv was 0. Hence, we assume that the bogus state due to the bug has been corrected.

PR #64, opened 2024-08-16
[AAA] Change mint stat calculations for payment type

DidacSF:
## Description

Changes stat calculations for minting so that free mints and normal mints got to different sections, instead of free mints also counting as normal mints.

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-awesome-avatars/src/tests.rs
+++ b/pallets/ajuna-awesome-avatars/src/tests.rs
@@ -1067,7 +1067,7 @@ mod minting {
 					let mut expected_nonce = 0;
 					let mut owned_avatar_count = 0;
 					let mut season_minted_count = 0;
-					let mut minted_count = 0;
+					let mut season_free_minted_count = 0;
 
 					let season_id = 1;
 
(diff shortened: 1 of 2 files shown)


pallets/ajuna-awesome-avatars/src/tests.rs
clangenb, line 1: Looks good to me, but I just think that this test is very long and hard to follow. So I have one question and one comment.

1. Do we expect different behavior here for different mint batch sizes?
2. I would recommend rewriting this test with [rstest](https://crates.io/crates/rstest) and parametrize this test to make it more readable and reduce the amount of written code.

clangenb (approved):
Looks good to me! I just have a comment and a question which is not directly related to the changes in this PR!

pallets/ajuna-awesome-avatars/src/tests.rs
DidacSF: 1. Not really
2. I would prefer not to introduce dependencies for testing

darkfriend77 (approved):
okay

PR #77, opened 2024-10-18
Sage Proto Type

clangenb:
This is a prototype of the sage engine focusing on how the downstream implementor would write their own state transition code. Hence, the internals of the sage engine have been neglected, but I made sure that the proposed design can be implemented using rust and frame, which is surprisingly difficult due to all the generics.

The high level overview is as follows:
1. Introduce a trait called `SageApi`, which implements everything we need.
2. Introduce a trait `SageGameTransition` with an associated type `SageApi`. The `SageGameTransition` has access to all the sage features via its associated typ `SageApi`.
3. Introduce a pallet `Sage`, which has two important associated types: `SageApi` and `SageGameTransition`. It doesn't really need the `SageApi`, but it is needed to constrain the associated types of our `SageGameTransition` and the `SageApi` to the runtime types (Balance, AccountId, etc).

This PR implements a tiny `ExampleGameTransition` and the sage pallets mock runtime shows how it would be integrated into an actual runtim.


sage/example-transition/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+	type Asset = Asset;
+	type AccountId = AccountId;
+	type Balance = Balance;
+
+	type TransitionId = ExampleTransitionId;
+	type Extra = ();
+
+	fn verify_rule<
+		Sage: SageApi<Asset = Self::Asset, AccountId = Self::AccountId, Balance = Self::Balance>,
+	>(
```
clangenb, line 62: I am not really happy with this yet, but the problem is that it is hard in rust to combine generic code with non-generic code. So we will never get around constraining those associated types.

I am wondering if we should make the sage pallet less generic in general, but using the explicit runtime types that we are using in the ajuna runtime.

So the steps would be:
1. create a crate `ajuna_runtime_primitives`
2. make `SageApi` not generic with associated types, but us our runtime types here.

However, I am not 100% sure if this is a good idea. I will think some more.

clangenb: The reason why I am suggesting this is: our approach here, should make it as easy as possible for downstream developers to deploy a game, so ideally they should not care about generics too much, but currently, this is hard to avoid.

clangenb: But maybe I am overthinking and adding 

```rust
Sage: SageApi<Asset = Self::Asset, AccountId = Self::AccountId, Balance = Self::Balance>,
```
to every function that the game studio writes and needs, the `SageApi` is not too bad after all.

sage/api/src/traits.rs
```diff
@@ -0,0 +1,71 @@
...
+	/// method. If you don't need custom arguments, you can define that type as `()`.
+	type Extra: Member + Encode + Decode + MaxEncodedLen + TypeInfo;
+
+	fn verify_rule<
+		Sage: SageApi<Asset = Self::Asset, AccountId = Self::AccountId, Balance = Self::Balance>,
+	>(
+		transition_id: Self::TransitionId,
+		account_id: &Self::AccountId,
+		assets: &[Self::Asset],
+		extra: &Self::Extra,
```
clangenb, line 69: I am considering splitting this up into `verify_rule` and `verify_rule_with_extra`, so that only games that use the `Extra` need to pass that argument.

.github/workflows/check-pull-request.yml
```diff
@@ -65,18 +65,6 @@ jobs:
...
-        pallet:
-          - pallet-ajuna-awesome-avatars
-          - pallet-ajuna-awesome-avatars-benchmarking
-          - pallet-ajuna-battle-mogs
-          - pallet-ajuna-nft-staking
-          - pallet-ajuna-nft-staking-benchmarking
-          - pallet-ajuna-nft-transfer
-          - pallet-ajuna-affiliates
-          - pallet-ajuna-tournament
-          - pallet-ajuna-wildcard
```
clangenb, line 79: I was just lazy in the beginning, but then I thought, why do we have it like that? With this pattern we have to add every crate in the workspace, not only the pallets. I think it is very easy to accidentally miss a crate, and then not test it. Hence, I would suggest running the tests on the workspace if you have nothing against that.

DidacSF: I agree that it makes more sense to do it like this

sage/api/src/traits.rs
DidacSF: hmm maybe its better to instead create an additional trait like `SageGameTransitionWithExtra` that contains that `verify_rule_with_extra` method and also the `Extra` associate type?

DidacSF:
I like what I see! For the more final version we should use and expand the AccountManager and AssetManager traits with either additional versions that can mutate state or just add the extra features on them so that we can easily use them in the SageTransition logic

sage/api/src/traits.rs
clangenb: I see what you mean, but then we would have to add another version of our sage pallet too because this `Extra` is also an extrinsic parameter.

DidacSF: Then in that case I think that for now its is fine the way it is

PR #80, opened 2024-10-25
[AJUN/730/772/774] Extract AAA code into Affiliates

DidacSF:
## Description

This PR extracts tries to decouple the logic of the Affiliates pallet from AAA as much as possible.

## Type of changes

## Checklist


pallets/ajuna-awesome-avatars/src/tests.rs
clangenb, line 1: I checked that all tests have been moved to the affiliators pallet, nice. 👍 But I see that there are still some affiliate tests in this file, shouldn't they be moved too?

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
```diff
@@ -0,0 +1,83 @@
...
+pub struct AffiliateUnlockParams<AccountId> {
+	pub target: UnlockTarget<AccountId>,
+	pub season_id: SeasonId,
+}
+
+impl<T: Config> AffiliateUnlockRules for Pallet<T> {
+	type AccountId = AccountIdFor<T>;
+	type UnlockParameters = AffiliateUnlockParams<Self::AccountId>;
+
+	fn try_validate_unlock(
```
clangenb: I think the naming is off here. Validate would imply to me that stuff is checked, not changed. Here, we mutate stuff and we even charge fees. Shall we rename it to  `unlock` or  `unlock_charging`?

Sidenote: I personally prefer to omit the prefix `try_` if there is not the same function without the prefix. I remember that this is the kind of the consensus in rust, as the result already makes it obvious that the function can fail.

pallets/ajuna-awesome-avatars/src/lib.rs
clangenb, line 1: * Shouldn't we move the `force_set_affiliation_chain` too?

* Also, rule mutator is no longer needed as a trait bound 🥳

pallets/ajuna-affiliates/src/mock.rs
```diff
@@ -44,6 +57,8 @@ frame_support::construct_runtime!(
 		Balances: pallet_balances = 1,
 		AffiliatesAlpha: pallet_ajuna_affiliates::<Instance1> = 2,
 		AffiliatesBeta: pallet_ajuna_affiliates::<Instance2> = 3,
+		#[cfg(feature = "runtime-benchmarks")]
+		AffiliatesBench: pallet_ajuna_affiliates = 4,
```
clangenb, line 50: What is the idea behind this?

pallets/ajuna-affiliates/src/mock.rs
```diff
@@ -31,10 +39,15 @@ pub type MockAccountId = <MockAccountPublic as IdentifyAccount>::AccountId;
...
+#[cfg(test)]
 pub const ALICE: MockAccountId = 1;
+#[cfg(test)]
 pub const BOB: MockAccountId = 2;
+#[cfg(test)]
 pub const CHARLIE: MockAccountId = 3;
+#[cfg(test)]
 pub const DAVE: MockAccountId = 4;
+#[cfg(test)]
 pub const EDWARD: MockAccountId = 5;
```
clangenb: These accounts are only needed in the tests. Instead of conditionally compiling them here, I suggest moving them to `test.rs`.

pallets/ajuna-affiliates/src/mock.rs
```diff
@@ -103,36 +118,177 @@ parameter_types! {
...
 	type RuntimeRule = MockRuntimeRule;
 	type AffiliateMaxLevel = AffiliateMaxLevel;
+	type UnlockParameters = MockUnlockParameter;
+	type AffiliatesUnlockRules = MockAffiliateRules;
+	type WeightInfo = ();
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = AffiliateBenchmarkHelper;
 }
 
+#[cfg(test)]
```
clangenb, line 256: I also suggest moving this to the test file, but I know that this is opinionated.

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
```diff
@@ -0,0 +1,83 @@
...
+		account: &Self::AccountId,
+		params: Self::UnlockParameters,
+	) -> Result<Self::AccountId, DispatchError> {
```
clangenb: I wouldn't return the account here, it introduces an unnecessary clone.

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
```diff
@@ -0,0 +1,83 @@
...
+						PlayerSeasonConfigs::<T>::mutate(account, season_id, |config| {
+							config.locks.affiliate = true;
+						});
+						Ok(account.clone())
+					} else {
+						Err(Error::<T>::UnlockCriteriaNotFulfilled.into())
+					}
+				} else {
+					Err(Error::<T>::FeatureLockedInSeason.into())
+				}
```
clangenb: I generally prefer if the code is more linear as it is easier to understand:

```rust
let affiliate_unlock = SeasonUnlocks::<T>::get(season_id)
   .map(|config| config.affiliate_unlock)
   .ok_or(Error::<T>::FeatureLockedInSeason.into())?;

let player_stats = SeasonStats::<T>::get(season_id, account);

if Self::evaluate_unlock_state(&unlock_vec, &player_stats) {
   PlayerSeasonConfigs::<T>::mutate(account, season_id, |config| {
	config.locks.affiliate = true;
   });
   Ok(account.clone())
} else {
   Err(Error::<T>::UnlockCriteriaNotFulfilled.into())
}
```

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
```diff
@@ -0,0 +1,83 @@
...
+						let GlobalConfig { affiliate_config, .. } = GlobalConfigs::<T>::get();
+						ensure!(
+							affiliate_config.affiliator_enable_fee > 0_u32.into(),
+							Error::<T>::FeatureLockedThroughPayment
+						);
+						T::Currency::transfer(
+							account,
+							&Self::treasury_account_id(),
+							affiliate_config.affiliator_enable_fee,
+							AllowDeath,
```
clangenb: I would recommend disallowing death here. A player might not like it, if the account is suddenly reaped.

clangenb (commented):
Nice, I only have a few remarks, otherwise, I think we are on the good track!

pallets/ajuna-awesome-avatars/src/tests.rs
DidacSF: Those tests make use of logic implemented in the AAA side, that's why they remain there for now, its not ideal true, but for now I didn't want to touch them.

pallets/ajuna-awesome-avatars/src/lib.rs
DidacSF: I didn't move it, because that method was designed as a patch to fix an issue more than a feature we would want to provide.

pallets/ajuna-awesome-avatars/src/tests.rs
clangenb: ok, we can leave this for later than, but there is one test, which has the #ignore tag, what about this one?

pallets/ajuna-affiliates/src/mock.rs
DidacSF: Make the benchmarks compile. It didn't manage to do it in a different way

pallets/ajuna-awesome-avatars/src/tests.rs
DidacSF: Hm, that was and old test that I forgot to un-ingore, will do for next commit

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
DidacSF: Renamed to `execute_unlock_rule_for`

pallets/ajuna-awesome-avatars/src/impls/affiliate_unlock_rules.rs
DidacSF: Adjusted so that account is not returned, changed type of `enable_affiliator` extrinsic

pallets/ajuna-affiliates/src/mock.rs
clangenb: Hmm, I see. I have looked into it a too, and could find a way. I guess it makes sense that they expect the test runtime to use the plain pallet instance.

pallets/ajuna-awesome-avatars/src/lib.rs
clangenb: hmm, I still think that we should either remove it or move it, but keeping it here doesn't make sense to me tbh.

DidacSF: Okay, I removed the extrinsic from AAA, we can later decide on what to do with it during later SAGE development

clangenb: Ok, cool thanks for putting up with me! 🙇‍♂️ 😄

clangenb (approved):
Nice, looks good to me!

PR #84, opened 2024-11-01
[AJUN-737/808] FeeHandler implementation

### LINKED ISSUE #83: Extract `GameFeeHandler` handler, which handles tournament and affiliate fees and treasury.
Instead of having the AAA pallet charging the fees, we want the fee handler to be responsible for that. The goal is that AAA only decides, when it wants to call the fee handler, but it should no longer handle the internals itself.

In a first step, it can be quite simple as it only handles the affiliate fee propagation and the rest into the treasury. It could look something like:

```rust
// some rust pseudocode

pub struct GameFeeHandler<AffiliatesInspector, TournamentInspector, TreasuryAccountId> {\
   ...
};

pub trait HandleGameFee {
 
    fn propagate_chain_fee(fee: Fee) -> Result<RemainingFees>;

    // not sure if a fee argument might make sense here
    fn charge_tournament_fee() -> Result<RemainingFee>;

    fn deposit_into_treasury(fee: Fee) -> Result<(), ...>; 

   // maybe we also want to expose a method like:
  fn charge_fee(fee: Fee) -> Result {
      let remaining = Self::propagate_chain_fee(fee)?;
      Self::deposit_into_treasury()
  }
    
}

// implement the trait for all generics, so that we can simply aggregate our fee handler in the runtime. 
// TBD: we might want to implement small subtraits something like:
// `AffiliateFeePropagation` or `ChargeTournamentFee`
impl<A: AffiliatInspector, T: TournamentInspector, TA: AccountId> for GameFeeHandler<A, T, TA> {
   ....
}

// If we don't want to charge for tournaments for instance, we could implement this in a 
// very simple way
impl<A: AffiliatInspector, T: TournamentInspector, TA: AccountId> for GameFeeHandler<A, T, TA> {
   ...

   fn charge_tournament_fee() -> Result {
       Ok(0);
   }
}
```

Maybe there is also some inspiration over here: https://github.com/paritytech/polkadot-sdk/blob/5f782a4c5836f9bd8f279e8dd8d3a0d03a88a64f/substrate/bin/node/runtime/src/lib.rs#L193

DidacSF:
## Description

This PR deals with the FeeHandler implementation described in #83 

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-affiliates/src/impls/fee_handler.rs
+++ b/pallets/ajuna-affiliates/src/impls/fee_handler.rs
@@ -1,0 +1,12 @@
+use super::*;
+use ajuna_primitives::fee_handler::FeeHandler;
+
+impl<T: Config<I>, I: 'static> FeeHandler for Pallet<T, I> {
+	type AccountId = AccountIdFor<T>;
+	type FeeIdentifier = RuleIdentifierFor<T, I>;
+	type FeeCurrency = BalanceOf<T, I>;
+
+	fn get_fee(account: &Self::AccountId, identifier: &Self::FeeIdentifier) -> Self::FeeCurrency {
+		todo!()
+	}
+}
--- a/pallets/ajuna-affiliates/src/impls/fee_handler.rs
+++ b/pallets/ajuna-affiliates/src/impls/fee_handler.rs
@@ -1,12 +1,35 @@
 use super::*;
-use ajuna_primitives::fee_handler::FeeHandler;
+use ajuna_primitives::fee_handler::FeeProvider;
+use sp_runtime::{traits::CheckedDiv, Saturating};
 
-impl<T: Config<I>, I: 'static> FeeHandler for Pallet<T, I> {
+impl<T: Config<I>, I: 'static> FeeProvider for Pallet<T, I> {
 	type AccountId = AccountIdFor<T>;
 	type FeeIdentifier = RuleIdentifierFor<T, I>;
 	type FeeCurrency = BalanceOf<T, I>;
+	type FeeOutput = Vec<(Self::FeeCurrency, Self::AccountId)>;
 
-	fn get_fee(account: &Self::AccountId, identifier: &Self::FeeIdentifier) -> Self::FeeCurrency {
-		todo!()
+	fn get_fee_from(
+		base_fee: Self::FeeCurrency,
+		account: &Self::AccountId,
+		identifier: &Self::FeeIdentifier,
+	) -> Self::FeeOutput {
+		if let Some(chain) = Self::get_affiliator_chain_for(account) {
+			if let Some(rule_chain) = Self::get_rule_for(identifier) {
+				rule_chain
+					.into_iter()
+					.map(|rule_perc| {
+						base_fee
+							.saturating_mul(rule_perc.into())
+							.checked_div(&100_u32.into())
+							.unwrap_or_default()
+					})
+					.zip(chain)
+					.collect()
+			} else {
+				Vec::with_capacity(0)
+			}
+		} else {
+			Vec::with_capacity(0)
+		}
 	}
 }
--- a/pallets/ajuna-awesome-avatars/src/lib.rs
+++ b/pallets/ajuna-awesome-avatars/src/lib.rs
@@ -742,7 +742,11 @@ pub mod pallet {
 				);
 
 				if affiliate_config.mode == AffiliateMode::Open && affiliate_config.enabled_in_buy {
-					Self::try_propagate_chain_fee(AffiliateMethods::Buy, &buyer, base_fee)?
+					T::FeeHandler::try_propagate_chain_fee(
+						base_fee,
+						&buyer,
+						&AffiliateMethods::Buy,
+					)?
 				} else {
 					base_fee
 				}
@@ -806,10 +810,10 @@ pub mod pallet {
 				if affiliate_config.mode == AffiliateMode::Open &&
 					affiliate_config.enabled_in_upgrade
 				{
-					Self::try_propagate_chain_fee(
-						AffiliateMethods::UpgradeStorage,
-						&caller,
+					T::FeeHandler::try_propagate_chain_fee(
 						base_fee,
+						&caller,
+						&AffiliateMethods::UpgradeStorage,
 					)?
 				} else {
 					base_fee
@@ -1312,27 +1316,21 @@ pub mod pallet {
 					let mint_fee = {
 						let base_fee = season.fee.mint.fee_for(&mint_option.pack_size);
 
-						let updated_fee =
-							match T::TournamentHandler::get_active_tournament_config_for(&season_id)
-							{
-								Some((
-									_,
-									TournamentConfig {
-										take_fee_percentage: Some(fee_perc), ..
-									},
-								)) if is_tournament_in_active_period => Self::try_propagate_tournament_fee(
-									&season_id, player, fee_perc, base_fee,
-								)?,
-								_ => base_fee,
-							};
+						let updated_fee = if is_tournament_in_active_period {
+							T::FeeHandler::try_propagate_tournament_fee(
+								base_fee, player, &season_id,
+							)?
+						} else {
+							base_fee
+						};
 
 						if affiliate_config.mode == AffiliateMode::Open &&
 							affiliate_config.enabled_in_mint
 						{
-							Self::try_propagate_chain_fee(
-								AffiliateMethods::Mint,
-								player,
+							T::FeeHandler::try_propagate_chain_fee(
 								updated_fee,
+								player,
+								&AffiliateMethods::Mint,
 							)?
 						} else {
 							updated_fee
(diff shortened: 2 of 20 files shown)


primitives/src/fee_handler.rs
```diff
@@ -0,0 +1,119 @@
...
+		base_fee: Self::FeeCurrency,
+		account: &Self::AccountId,
+		identifier: &Self::AffiliateFeeIdentifier,
+	) -> Result<Self::FeeCurrency, DispatchError> {
+		let mut final_fee = base_fee;
+
+		for (transfer_fee, chain_account) in Affiliate::get_fee_from(base_fee, account, identifier)
+		{
+			if transfer_fee > 0_u32.into() {
+				CurrencyHandler::transfer(account, &chain_account, transfer_fee, AllowDeath)?;
```
clangenb: Again, `AllowDeath` should not be the choise here, better keep alive.

primitives/src/fee_handler.rs
```diff
@@ -0,0 +1,119 @@
...
+				final_fee = final_fee.saturating_sub(transfer_fee);
```
clangenb: Just do be sure, I think we should do a checked_sub here, and return in case of an error. What do you think?

primitives/src/fee_handler.rs
```diff
@@ -0,0 +1,119 @@
...
+	fn try_propagate_tournament_fee(
+		base_fee: Self::FeeCurrency,
+		account: &Self::AccountId,
+		identifier: &Self::TournamentFeeIdentifier,
+	) -> Result<Self::FeeCurrency, DispatchError> {
+		let (tournament_fee, tournament_account) =
+			Tournament::get_fee_from(base_fee, account, identifier);
+
+		if tournament_fee > 0_u32.into() {
+			CurrencyHandler::transfer(account, &tournament_account, tournament_fee, AllowDeath)?;
```
clangenb: ditto

primitives/src/fee_handler.rs
```diff
@@ -0,0 +1,119 @@
...
+			Ok(base_fee.saturating_sub(tournament_fee))
```
clangenb: Just to be sure, I think I would prefer to return an error here too.

primitives/src/treasury_manager.rs
```diff
@@ -0,0 +1,40 @@
...
+use parity_scale_codec::Codec;
+
+/// The treasury manager trait that can be passed around to other pallets that need to works with
+/// the treasury or its accounts
+pub trait TreasuryManager {
+	type AccountId: Member + Codec;
+
+	type Currency: Member + Codec;
+
+	type TreasuryKey: Parameter + Member;
```
clangenb: I think we should call it just `TreasuryAccount`

primitives/src/treasury_manager.rs
```diff
@@ -0,0 +1,40 @@
...
+
+	fn is_treasurer(account: &Self::AccountId) -> Result<(), DispatchError>;
+
+	fn get_treasurer() -> Result<Self::AccountId, DispatchError>;
+
+	#[cfg(feature = "runtime-benchmarks")]
+	fn set_treasurer(owner: Self::AccountId);
```
clangenb: May I ask for some documentation about treasury vs. treasurer, it is not obvious to me.

pallets/ajuna-affiliates/src/impls/fee_handler.rs
```diff
@@ -0,0 +1,35 @@
...
+					.zip(chain)
+					.collect()
+			} else {
+				Vec::with_capacity(0)
+			}
+		} else {
+			Vec::with_capacity(0)
+		}
+	}
+}
```
clangenb, line 35: I am wondering if we should make this trait not depend on the balance type, but only return the fractions of the fee propagation. If we do it like that, the pallet will be independent of the currency. It is then the fee handlers responsibility to translate this in to actual balance values. What do you think about that?

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -735,7 +742,11 @@ pub mod pallet {
 				);
 
 				if affiliate_config.mode == AffiliateMode::Open && affiliate_config.enabled_in_buy {
```
clangenb, line 744: If we think about this from the sage context, I think it is the fee handler itself, which should determine if it is enabled or not, but maybe we do it in a second step. It is not obvious to me now, how we best implement this, as all the config stuff is in the AAA. What do you think, could we move this config stuff into the affiliates-pallet?

pallets/ajuna-tournament/src/impls/fee_handler.rs
```diff
@@ -0,0 +1,37 @@
...
+use sp_arithmetic::traits::{CheckedDiv, Saturating};
+
+impl<T: Config<I>, I: 'static> FeeProvider for Pallet<T, I> {
+	type AccountId = AccountIdFor<T>;
+	type FeeIdentifier = TournamentCategoryIdFor<T, I>;
+	type FeeCurrency = BalanceOf<T, I>;
+	type FeeOutput = (BalanceOf<T, I>, Self::AccountId);
+
+	fn get_fee_from(
+		base_fee: Self::FeeCurrency,
```
clangenb, line 12: Same comment as in the other implementation. Do we maybe just want to return the fractions here?

pallets/ajuna-tournament/src/impls/fee_handler.rs
```diff
@@ -0,0 +1,37 @@
...
+		let tournament_account = Self::get_treasury_account_for(identifier);
+
+		match Self::get_active_tournament_config_for(identifier) {
+			Some((_, TournamentConfig { take_fee_percentage: Some(fee_perc), .. }))
+				if is_tournament_in_active_period =>
+			{
+				let tournament_fee = base_fee
+					.saturating_mul(fee_perc.into())
+					.checked_div(&100_u32.into())
+					.unwrap_or_default();
```
clangenb, line 30: I think it could be easier, like:

```rust
let tournament_fee = Permill::from_percent(fee_perc).saturating_mul(base_fee);
```

We could even decide to use the Permill type in the storage, but if we set it as an extrinsic, we just have to be aware that we are using per millions instead of per hundreds. I guess this is too error prone.

clangenb (commented):
All in all this looks very close to what I imagined, and the differences are mostly justified I think!

I just have some minor questions and comments.

primitives/src/treasury_manager.rs
DidacSF: It's not exactly an account, the key is just a way to identify a given treasury pot, since we do have some cases in which we split the treasury based on criteria like the season_id for example

pallets/ajuna-affiliates/src/impls/fee_handler.rs
DidacSF: I think this approach is better since it is more clear who is responsible for calculating the fees, additionally with this approach we can indicate if no fee is to be paid for something wheres with the fractions you would still need to compute if the fraction leads to paying any fees or not at the outer level.

pallets/ajuna-awesome-avatars/src/lib.rs
DidacSF: I think the best way to make this independent is to, much like we did for other components, define a trait or set of traits that allow the inspection of the configuration, that way the sage-pallet only needs to implement the traits and become the configuration provider, or we could even define a separate pallet just to hold the configuration if we find that better.

TLDR: Yes, this should be discussed in the second round of development

pallets/ajuna-tournament/src/impls/fee_handler.rs
DidacSF: Just as in the other comment, I think its better to actually make the calculation at the implementer side.

pallets/ajuna-tournament/src/impls/fee_handler.rs
DidacSF: Hmm, I think its more clear the way it is expressed right now, as you say. Doing it per million is less familiar to most people I think.

pallets/ajuna-awesome-avatars/src/lib.rs
clangenb: Ok, cool, I agree. I generally wonder, if global config needs to be defined aggregated in the core pallet or if it would be much better to configure the pallet sovereignly in this case too.

DidacSF: I think that all config needed to support the features the core pallet has should be there, but we need to think on which config has to be shared and how customizable that configuration can be in terms of types and generics.

pallets/ajuna-affiliates/src/impls/fee_handler.rs
clangenb: Yes, I see your point. However, my arguments would be:

1. It would make sense to me if the instance computing the actual fee is the fee handler, instead of the pallet itself.
2. The fee handler could also early return if the base fee is 0 and doesn't even call into this code, which is more intuitive to me instead of first having to look into the affiliate code, and then understand the code, and then see, ah, it just returns a `Vec<(0, Beneficifiary>`.

I still tend towards moving the fee calculation into the fee handler for better separation of concerns, IMO the pallet should not care about how the distribution is interpreted, but we can also do this in a second step. 👍

Now, that I write this, I think we should definitely do the early return anyhow for readability and efficiency :D. Hence, implement my suggestion from 2.

DidacSF: Hm, I agree to look into this again in step 2

primitives/src/treasury_manager.rs
clangenb: So, can we maybe say that it is a `SubTreasuryKey` if I understand you correctly?

pallets/ajuna-awesome-avatars/src/lib.rs
clangenb: Exactly, I agree with what you say. We should revisit, which things are actually also needed by the core pallet itself, and if yes, it may obviously stay there.

clangenb: but I do think, again, that with the right abstractions we can move some stuff out of the core pallet.

primitives/src/treasury_manager.rs
DidacSF: Hmm maybe more like `TreasuryPotKey`?

clangenb (approved):
Looks good to me! Thanks for fixing my nitpicks. There is only one more question about the `TreasuryKey`, but I can already approve it.

If you agree though that it is the subtreasury id, I think we should rename the the associated type though.

PR #85, opened 2024-11-11
[AJUN-811] Implement pallet-sage core

DidacSF:
## Description

This PR is the first full implementation of the pallet-sage with full featureset

## Type of changes

## Checklist


sage/pallet/src/config/player.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Default, Debug, PartialEq)]
+pub struct PlayerStats<BlockNumber> {
+	pub minted_amount: Stat,
+	pub forged_amount: Stat,
+	pub bought_amount: Stat,
+	pub sold_amount: Stat,
+	pub first_mint: Option<BlockNumber>,
+	pub latest_mint: Option<BlockNumber>,
+	pub first_forge: Option<BlockNumber>,
+	pub latest_forge: Option<BlockNumber>,
```
clangenb, line 85: Maybe we can change this into a transition stat: `transitions_executed: Map<Id, Stat>`

sage/pallet/src/lib.rs
```diff
@@ -62,13 +93,113 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub type PlayerSeasonStats<T: Config<I>, I: 'static = ()> = StorageDoubleMap<
+		_,
+		Identity,
+		AccountIdOf<T>,
+		Identity,
+		SeasonIdOf<T, I>,
+		PlayerStats<BlockNumberFor<T>>,
+		ValueQuery,
+	>;
```
clangenb: Could we move this to the season handler too?

sage/pallet/src/lib.rs
```diff
@@ -113,37 +557,49 @@ pub mod pallet {
 		}
 	}
 
-	/// Implement the SageApi for this instance, this is essentially where all the associated
-	/// types are wired together and aggregated to the API.
-	impl<T: Config<I>, I: 'static> SageApi for Pallet<T, I> {
-		type AssetId = AssetIdOf<T, I>;
```
clangenb, line 119: I wanted to remove this implementation because

```rust
impl crate::Config<SageExampleTransitionInstance> for Test {
	...
	type SageGameTransition = ExampleTransitionGeneric<MockAccountId, MockBalance, pallet_sage::Pallet<Runtime>>;
	type SageApi = pallet_sage::Pallet<Runtime>>;
}
```

Here we would run into the recursion error. There is a way around this, but I have not fully evaluated the trade-offs yet. I would say the trade-off is: boilerplate code vs. flexibility mostly though.

sage/pallet/src/pallet_impls/unlocks.rs
```diff
@@ -0,0 +1,173 @@
...
+						ensure!(
+								fee.avatar_transfer_unlock > 0_u32.into(),
+								Error::<T, I>::FeatureLockedThroughPayment
+							);
+						T::Currency::transfer(
+							&account,
+							&Self::treasury_account_id(),
+							fee.avatar_transfer_unlock,
+							AllowDeath,
+						)?;*/
```
clangenb: Having a smooth abstraction here, will be one of the core issues we have to tackle in the sage architecture design, but I am confident that we can do it.

sage/pallet/src/pallet_impls/ensures.rs
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	pub(crate) fn ensure_can_be_set_for_trade(_asset_id: &AssetIdOf<T, I>) -> DispatchResult {
+		// TODO: Need to implement this either as TradeManager or something similar
+		// sine we need the concrete Asset type to apply any sort of filtering logic to it
+		// maybe expand the AssetT type
+
+		/*let trade_filters = SeasonTradeFilters::<T>::get(Asset.season_id)
+			.ok_or::<DispatchError>(Error::<T, I>::UnknownSeason.into())?;
+		ensure!(trade_filters.is_tradable(Asset), Error::<T, I>::AssetCannotBeTraded);*/
```
clangenb: I agree with you, and instead of an implicit filter of the DNA I would prefer to have an explicit `filter` method in the `Asset` trait. The concrete asset implementation can still choose if it internally wants to derive the filter from the DNA or even have an extra field.

clangenb (commented):
In general, it looks good. I just have a few comments, but I guess all the critical things is now what follows.

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -1241,9 +1238,15 @@ pub mod pallet {
...
 		}
 
-		pub(crate) fn deposit_into_treasury(season_id: &SeasonId, amount: BalanceOf<T>) {
+		pub(crate) fn deposit_into_treasury(
+			who: &AccountIdFor<T>,
+			season_id: &SeasonId,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			T::Currency::transfer(who, &Self::treasury_account_id(), amount, AllowDeath)?;
+			// Deal with the fact the account may not receive enough funds to be created
```
clangenb: As this is the treasury account, we can expect it to exist a priori anyhow. So this is fine.

sage/pallet/src/config/affiliates.rs
```diff
@@ -0,0 +1,24 @@
...
+// GNU Affero General Public License for more details.
+
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use frame_support::pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo};
+
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Debug, PartialEq, Eq)]
+pub enum AffiliateMethods {
+	StateTransition,
```
clangenb: As far as I understand, this should ideally be more generic, so that we can decide which state transitions are eligible for state transition. So I believe, we want to have an associated type `ElligibleMethods`, which could look like:

```rust
pub enum AffiliateMethods<TransitionId> {
     UpgradeAssetInventory,
     TradeAsset
     Transition {
          id: TransitionId
     }
}
```

sage/pallet/src/lib.rs
```diff
@@ -39,19 +61,51 @@ pub mod pallet {
...
+		>;
+
+		type FeeHandler: FeeHandler<
+			AccountId = AccountIdOf<Self>,
+			FeeCurrency = BalanceOf<Self, I>,
+			AffiliateFeeIdentifier = AffiliateMethods,
+			TournamentFeeIdentifier = SeasonIdOf<Self, I>,
+			// TODO: Define if we want this as a configurable parameter or some different fixed
+			// value
+			TreasuryKey = SeasonIdOf<Self, I>,
```
clangenb, line 108: TBD indeed, not sure yet.

sage/pallet/src/lib.rs
```diff
@@ -85,9 +290,284 @@ pub mod pallet {
...
+			let (seller, price) = Self::ensure_for_trade(&asset_id)?;
+			ensure!(buyer != seller, Error::<T, I>::AlreadyOwned);
+			// TODO: Should this segment be handled by the FeeHandler?
+			T::Currency::transfer(
+				&buyer,
+				&seller,
+				price,
+				frame_support::traits::ExistenceRequirement::KeepAlive,
+			)?;
+
```
clangenb: I think this one can remain here, not the fee handler, as it is essentially not the fee that is being charged. Instead, we transfer the money.

sage/pallet/src/lib.rs
```diff
@@ -85,9 +290,284 @@ pub mod pallet {
...
+				let percentage_fee = price.saturating_mul(fee.buy_percent.unique_saturated_into()) /
+					MAX_PERCENTAGE.unique_saturated_into();
+				let base_fee = sp_std::cmp::max(min_buy_fee, percentage_fee);
+
+				T::FeeHandler::try_propagate_chain_fee(
+					base_fee,
+					&buyer,
+					&AffiliateMethods::TradeAsset,
+				)?
+			};
```
clangenb, line 523: I think propagating the chain fee and depositing the leftovers into the treasury should be done in the same method, we will probably never want to configure this more granular.

sage/pallet/src/lib.rs
```diff
@@ -96,6 +576,12 @@ pub mod pallet {
 		) -> DispatchResult {
 			let sender = ensure_signed(origin)?;
 
+			// TODO: Maybe we should limit the maximum amount of asset_ids?
+			for asset_id in asset_ids.iter() {
+				Self::ensure_ownership(&sender, asset_id)?;
+				Self::ensure_unlocked(asset_id)?;
+			}
+
```
clangenb: Yes you are right, this is an attack vector. The argument should be a bounded vec with max length of the maximum ownership possible.

sage/pallet/src/lib.rs
```diff
@@ -107,43 +593,75 @@ pub mod pallet {
...
+				)?;
+				T::FeeHandler::try_propagate_chain_fee(
+					updated_fee,
+					&sender,
+					&AffiliateMethods::StateTransition,
+				)?
+			};
+			T::FeeHandler::deposit_fee_into_treasury(&sender, &current_season_id, transition_fee)?;
+
 			Self::deposit_event(Event::TransitionExecuted { account: sender, id: transition_id });
```
clangenb: If we play this wisely, we can remove the tournament dependency of the sage pallet. 

So, there is one primary question: Do we ever need the fee of the past season ID? If no, I think that we just need a `CurrentFee` in the tournaments-pallet, and we simply update it whenever we want. This would probably simplify the design a lot.

This makes me realize, we should question this for all config values: do we need to be able to index them by seasons? Or is mutating a current value enough? This might simplify so many interfaces big time, as we can omit the `SeasonId` parameter.

So, I have in mind for the interface here:

```rust

// the state transition can define if they want to charge anything and how much
let maybe fee = T::SageApi::execute_transition()?;

if let Some(fee) = maybe_fee { 

    // The fee handler can internally get the current season (if needed), and first
   // charge the tournament fee (if any), propagate the chain fee, and put the rest
   // into the treasury. What do you think?
    T::FeeHandler::handle_fee(fee, &sender, AffiliateMethods::StateTransition {id: t_id})?;
}
```

What I like about this idea, is that it is analogous to the substrate design.

sage/pallet/src/trait_impls/asset_manager.rs
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	fn handle_asset_prepare_fee(
+		_asset: &Self::Asset,
+		_player: &Self::AccountId,
+		_fee_recipient: &Self::AccountId,
+	) -> Result<(), DispatchError> {
+		todo!()
+	}
+
```
clangenb, line 110: Todo: move these to nft-transfer-pallet

sage/pallet/src/mock.rs
```diff
@@ -156,21 +171,179 @@ impl SageApi for SageMock {
...
 	type SageApi = SageMock;
+	type SeasonHandler = MockSeasonManager;
+	type FeeHandler = GameFeeHandler<
+		MockAccountId,
+		Balances,
+		MockAffiliatesFeeProvider,
+		MockTournamentFeeProvider,
+		MockTreasuryManager,
+	>;
+	type TradeHandler = MockTradeHandler;
```
clangenb, line 336: What kind of customizations do we expect here? I am not sure if we need this type.

sage/pallet/src/trait_impls/account_manager.rs
clangenb, line 1: These tests are relatively small, I wonder if it might be easier if they are in the same file as the implementation, what do you think?

clangenb (commented):
All in all I like it a lot actually! I haven't checked the tests yet, but I think it makes a lot of sense to have the extrinsic tests in a separate file.

Some things that definitely make sense discussing, but I think we are heading into a good direction.

sage/pallet/src/lib.rs
darkfriend77: So there is a few things to consider here, until now, and this indicates that we should be open for change, until now, a season technically never ended in terms of playing it, why? Because the only thing that the season limited was the minting transition, the forging of a season never stopped, and so did the trading of the season avatars.

I think this creates the next question that arise from the transitions side, because transitions in fact have a life cycle that can be limited by the season. Which then should answer the fee question as the fees are bound to the transition.

sage/pallet/src/lib.rs
```diff
@@ -226,9 +226,9 @@ pub mod pallet {
 			price: BalanceOf<T, I>,
 		},
 		/// Asset locked.
-		AssetLocked { asset_id: AssetIdOf<T, I> },
+		AssetLocked { asset_id: AssetIdOf<T, I>, lock: Lock<AccountIdOf<T>> },
 		/// Asset unlocked.
-		AssetUnlocked { asset_id: AssetIdOf<T, I> },
+		AssetUnlocked { asset_id: AssetIdOf<T, I>, lock: Lock<AccountIdOf<T>> },
```
clangenb, line 235: I think this makes sense. 👍

sage/pallet/src/tests/extrinsics/lock_asset.rs
```diff
@@ -57,48 +61,3 @@ fn can_lock_asset_successfully() {
 			}
 		});
 }
-
-#[test]
-fn cannot_lock_unowned_asset() {
```
clangenb: What happened to these tests?

Edit: nevermind - I saw that they were moved to the asset_manager.

sage/pallet/src/tests/extrinsics/unlock_asset.rs
```diff
@@ -17,49 +17,22 @@
...
+			asset_id,
+			lock: expected_lock,
+		}));
 		assert_eq!(LockedAssets::<Test, Instance1>::get(asset_id), None);
-		System::assert_has_event(RuntimeEvent::Sage(Event::AssetUnlocked { asset_id }));
-	});
-}
-
-#[test]
-fn cannot_unlock_non_owned_asset() {
```
clangenb: Same here, why have these been removed?

sage/pallet/src/trait_impls/asset_manager.rs
```diff
@@ -56,8 +48,9 @@ impl<T: Config<I>, I: 'static> AssetManager for Pallet<T, I> {
 			Ok(())
 		})?;
 
-		LockedAssets::<T, I>::insert(&asset_id, Lock::new(lock_id, owner));
-		Self::deposit_event(Event::AssetLocked { asset_id });
+		let lock = Lock::new(lock_id, owner);
+		LockedAssets::<T, I>::insert(&asset_id, lock.clone());
```
clangenb: I think we can pass this by reference. So we can omit the clone

sage/pallet/src/mock.rs
```diff
@@ -124,19 +139,21 @@ impl SageApi for SageMock {
...
-		_account: &Self::AccountId,
-		_asset: &Self::AssetId,
+		account: &Self::AccountId,
+		asset: &Self::AssetId,
 	) -> Result<(), sage_api::Error> {
 		// this would be a call to our asset manager implementation
-		todo!()
+		Sage::ensure_ownership(account, asset)
+			.map(|_| ())
+			.map_err(|_| sage_api::Error::Transition { error: 0 })
```
clangenb: Wow, I thought that this would end up in a recursion error. I am surprised.

sage/pallet/src/tests/extrinsics/state_transition.rs
```diff
@@ -0,0 +1,115 @@
...
+			let transition_id = ExampleTransitionId::UpgradeAsset;
+
+			assert_noop!(
+				Sage::state_transition(RuntimeOrigin::signed(ALICE), transition_id, asset_ids, ()),
+				Error::<Test, Instance1>::RuleNotSatisfied {
+					code: sage_api::Error::InvalidAssetLength.as_error_code()
+				}
+			);
+		})
+}
```
clangenb, line 113: Nice, I like the showcase.

sage/pallet/src/trait_impls/sage_api.rs
```diff
@@ -0,0 +1,45 @@
+// This program is free software: you can redistribute it and/or modify
```
clangenb: I think we can delete this.

sage/pallet/src/pallet_impls/unlocks.rs
```diff
@@ -0,0 +1,197 @@
...
+		//let free_minted = u8::try_from(account_stats.free_minted).unwrap_or(u8::MAX);
+		let forged = u8::try_from(account_stats.forged_amount).unwrap_or(u8::MAX);
+		let bought = u8::try_from(account_stats.bought_amount).unwrap_or(u8::MAX);
+		let sold = u8::try_from(account_stats.sold_amount).unwrap_or(u8::MAX);
+
+		config[0] <= minted &&
+			//config[1] <= free_minted &&
+			config[2] <= forged &&
+			config[3] <= bought &&
+			config[4] <= sold
```
clangenb, line 199: I would like to add a test here. Theoretically, this can panic if we change the size of the UnlockRule vec, e.g. this panics if we replace `config[4]` with `config[6]`. I would like to change this function to return a result (preferred) or default to false if the array does not have the expected length.

sage/pallet/src/config/player.rs
DidacSF: TBD

sage/pallet/src/trait_impls/account_manager.rs
DidacSF: For now, I would prefer to keep everything separated even if it ends up being more complex in terms of file organization.

sage/pallet/src/mock.rs
DidacSF: We need it because the Asset's internal structure is not known to the SAGE pallet so we need an external source that actually knows about it to implement any filtering logic

PR #87, opened 2024-11-19
Minor code improvements

clangenb:
I went through the sage code with a more low-level look and directly added some changes instead of just giving a review.

### DIFF
--- a/sage/pallet/src/lib.rs
+++ b/sage/pallet/src/lib.rs
@@ -313,7 +313,9 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			new_config: GeneralConfigOf<T, I>,
 		) -> DispatchResult {
-			Self::ensure_organizer(origin)?;
+			let signer = ensure_signed(origin)?;
+			Self::ensure_organizer(&signer)?;
+
 			GeneralConfigStore::<T, I>::put(&new_config);
 			Self::deposit_event(Event::UpdatedGeneralConfig { updated_config: new_config });
 			Ok(())
@@ -327,7 +329,8 @@ pub mod pallet {
 			feature: LockableFeature,
 			unlock_rule: UnlockRule,
 		) -> DispatchResult {
-			Self::ensure_organizer(origin)?;
+			let signer = ensure_signed(origin)?;
+			Self::ensure_organizer(&signer)?;
 			SeasonUnlocks::<T, I>::mutate(&season_id, feature, |config| {
 				*config = Some(unlock_rule);
 			});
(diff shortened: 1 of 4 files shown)


sage/pallet/src/lib.rs
```diff
@@ -374,6 +365,12 @@ pub mod pallet {
 			};
 			T::FeeHandler::deposit_fee_into_treasury(&caller, &season_id, upgrade_fee)?;
 
+			let account_to_upgrade = beneficiary.unwrap_or_else(|| caller);
+
+			let inventory_tier =
+				PlayerSeasonConfigs::<T, I>::get(&account_to_upgrade, &season_id).inventory_tier;
+			ensure!(inventory_tier != InventoryTier::Max, Error::<T, I>::MaxStorageTierReached);
+
```
clangenb, line 373: If we move this past the fee propagation, we can omit the clone.

sage/pallet/src/lib.rs
```diff
@@ -313,7 +313,9 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			new_config: GeneralConfigOf<T, I>,
 		) -> DispatchResult {
-			Self::ensure_organizer(origin)?;
+			let signer = ensure_signed(origin)?;
+			Self::ensure_organizer(&signer)?;
```
clangenb, line 317: I know that we have to write two lines each time here, but in general, I believe that the code is more ergonomic if we make the `ensure_organizer` just take an account id. The benefit can be seen further below.

sage/pallet/src/lib.rs
```diff
@@ -395,14 +392,14 @@ pub mod pallet {
...
-				Ok(organizer) => organizer,
-				_ => {
-					ensure!(transfer.open, Error::<T, I>::TransferClosed);
-					ensure_signed(origin)?
-				},
-			};
+			ensure!(
+				Self::ensure_organizer(&from).is_ok() || transfer.open,
+				Error::<T, I>::TransferClosed
+			);
```
clangenb: Here, we simplify the code a lot and make it more efficient with the new `ensure_organizer`.

sage/pallet/src/tests/extrinsics/upgrade_asset_inventory.rs
```diff
@@ -220,14 +220,19 @@ fn upgrade_asset_inventory_should_reject_insufficient_balance() {
 
 #[test]
 fn upgrade_asset_inventory_should_reject_fully_upgraded_storage() {
-	ExtBuilder::default().build().execute_with(|| {
-		PlayerSeasonConfigs::<Test, Instance1>::mutate(ALICE, SEASON_ID_0, |config| {
-			config.inventory_tier = InventoryTier::Max
-		});
+	let alice_initial_balance = MockExistentialDeposit::get() * 10;
```
clangenb, line 223: Had to give alice some funds as we first propagate the fee, and then check for the current storage upgrade level.

PR #93, opened 2024-11-29
[AJUN-874] SAGE pallet/api refinement

DidacSF:
## Description

This PR refactors and refines some of the code associated with the SAGE pallet and its API surface traits. The changes are the following:

* Added the `Identifiable` trait which is required for `Asset` types to implement, used to generate `AssetId`
* Move `TransitionConfig` to `SageGameTransition` trait
* Add missing benchmarks to `pallet-sage-core`
* Remove all methods in `ajuna-primitives` traits that required `runtime-benchmarks`

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-nft-transfer/src/lib.rs
+++ b/pallets/ajuna-nft-transfer/src/lib.rs
@@ -62,13 +62,6 @@ pub mod pallet {
 		fn create_items(owner: AccountId, count: u32) -> sp_std::vec::Vec<ItemId>;
 	}
 
-	#[cfg(feature = "runtime-benchmarks")]
-	impl<AccountId, ItemId> BenchmarkHelper<AccountId, ItemId> for () {
-		fn create_items(_owner: AccountId, _count: u32) -> sp_std::vec::Vec<ItemId> {
-			sp_std::vec::Vec::with_capacity(0)
-		}
-	}
-
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
--- a/pallets/ajuna-nft-transfer/src/lib.rs
+++ b/pallets/ajuna-nft-transfer/src/lib.rs
@@ -57,6 +57,18 @@ pub mod pallet {
 		Uploaded,
 	}
 
+	#[cfg(feature = "runtime-benchmarks")]
+	pub trait BenchmarkHelper<AccountId, ItemId> {
+		fn create_items(owner: AccountId, count: u32) -> sp_std::vec::Vec<ItemId>;
+	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	impl<AccountId, ItemId> BenchmarkHelper<AccountId, ItemId> for () {
+		fn create_items(_owner: AccountId, _count: u32) -> sp_std::vec::Vec<ItemId> {
+			sp_std::vec::Vec::with_capacity(0)
+		}
+	}
+
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
@@ -101,6 +113,9 @@ pub mod pallet {
 			+ Mutate<Self::AccountId, Self::ItemConfig>;
 
 		type WeightInfo: WeightInfo;
+
+		#[cfg(feature = "runtime-benchmarks")]
+		type BenchmarkHelper: BenchmarkHelper<AccountIdFor<Self>, Self::ItemId>;
 	}
 
 	#[pallet::storage]
--- a/pallets/ajuna-tournament/src/lib.rs
+++ b/pallets/ajuna-tournament/src/lib.rs
@@ -84,41 +84,6 @@ pub mod pallet {
 		fn create_entities(owner: AccountId, count: u32) -> sp_std::vec::Vec<(EntityId, Entity)>;
 	}
 
-	#[cfg(feature = "runtime-benchmarks")]
-	impl<
-			CategoryId: From<u32>,
-			BlockNumber: From<u64>,
-			Balance: From<u64>,
-			Ranker: Default,
-			AccountId,
-			EntityId,
-			Entity,
-		> BenchmarkHelper<CategoryId, BlockNumber, Balance, Ranker, AccountId, EntityId, Entity> for ()
-	{
-		fn create_category_id(id: u32) -> CategoryId {
-			id.into()
-		}
-
-		fn create_default_tournament_config() -> TournamentConfig<BlockNumber, Balance, Ranker> {
-			TournamentConfig {
-				start: 20_u64.into(),
-				active_end: 40_u64.into(),
-				claim_end: 50_u64.into(),
-				initial_reward: None,
-				max_reward: None,
-				take_fee_percentage: None,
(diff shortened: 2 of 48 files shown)


pallets/ajuna-nft-transfer/src/lib.rs
```diff
@@ -57,6 +57,18 @@ pub mod pallet {
...
+	pub trait BenchmarkHelper<AccountId, ItemId> {
+		fn create_items(owner: AccountId, count: u32) -> sp_std::vec::Vec<ItemId>;
+	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	impl<AccountId, ItemId> BenchmarkHelper<AccountId, ItemId> for () {
+		fn create_items(_owner: AccountId, _count: u32) -> sp_std::vec::Vec<ItemId> {
+			sp_std::vec::Vec::with_capacity(0)
+		}
+	}
```
clangenb: The benchmark will fail if you use `()`. So I suggest dropping this implementation or make it meaningful.

sage/api/src/traits.rs
```diff
@@ -4,6 +4,10 @@ use scale_info::TypeInfo;
 use sp_runtime::traits::Member;
 use sp_std::vec::Vec;
 
+pub trait Identifiable<Id> {
+	fn get_id(&self) -> Id;
+}
+
```
clangenb: It is the consensus that traits exposing only one method should be called like that method, i.e. `GetId`.

sage/example-transition/src/generic.rs
```diff
@@ -50,19 +59,40 @@ where
...
+						base
+					};
+					let asset =
+						Asset::create(asset_id, i as u32, 0, 0, [i as u8; 32], 10, Level::Two);
+					output_vec.push(TransitionOutput::Minted(asset))
+				}
+
+				Ok(output_vec)
+			},
+		}
```
clangenb, line 95: What was your thinking behind that?

In my understanding, we don't really need a bench transition that is really expensive, but the concrete benchmark implementation in the runtime should just implement a benchmark helper that retrieves the id of the most expensive benchmark. What do you think?

sage/pallet/src/benchmarking.rs
```diff
@@ -0,0 +1,402 @@
...
+use frame_system::RawOrigin;
+use sp_runtime::BuildStorage;
+
+parameter_types! {
+	pub const BenchPalletId: PalletId = PalletId(*b"sage/bch");
+}
+
+pub struct MockAssetMediatorBench;
+
+impl AssetManager for MockAssetMediatorBench {
```
clangenb: Hmm, I start to think that our mock runtime shouldn't use instances in the first place. It is not necessary, and we can just omit the redundant implementation here.

clangenb (commented):
Some minor comments, but otherwise it looks good!

sage/example-transition/src/generic.rs
DidacSF: I agree with your comment, I'll adjust the benchmark helper so that it works like you mention

sage/pallet/src/benchmarking.rs
DidacSF: Okay, I've removed the Instance from the mock runtime and adapted the tests and benchmarks for it

sage/pallet/src/tests/extrinsics/unlock_feature.rs
```diff
@@ -216,10 +188,6 @@ fn unlock_feature_works_for_transfer_asset_targets() {
 				Balances::free_balance(BOB),
 				initial_balance - season_config.fee.unlock_trade_asset * 2
 			);
-			assert!(
-				PlayerSeasonConfigs::<Test, Instance1>::get(CHARLIE, season_id)
-					.locks
-					.asset_transfer
-			);
+			assert!(PlayerSeasonConfigs::<Test, ()>::get(CHARLIE, season_id).locks.asset_transfer);
```
clangenb, line 191: Awesome side effect XD

pallets/ajuna-tournament/src/lib.rs
```diff
@@ -103,6 +113,10 @@ pub mod pallet {
 				ranker: Ranker::default(),
 			}
 		}
+
+		fn create_entities(_owner: AccountId, _count: u32) -> sp_std::vec::Vec<(EntityId, Entity)> {
+			sp_std::vec::Vec::with_capacity(0)
+		}
```
clangenb: The benchmarks will also panic in this case, so I suggest removing this implementation.

sage/pallet/src/lib.rs
```diff
@@ -140,6 +176,19 @@ pub mod pallet {
...
+			AssetIdOf<Self, I>,
+			AssetOf<Self, I>,
+			TradeFilterOf<Self, I>,
+			TransferFilterOf<Self, I>,
+			AccountIdOf<Self>,
+			SeasonIdOf<Self, I>,
+			TransitionIdOf<Self, I>,
+			TransitionConfigOf<Self, I>,
+			ExtraOf<Self, I>,
+		>;
```
clangenb: Do we really need this benchmark helper? Is it not possible to (at least partly) just use the trait in the benchmarks?

My opinion is here that Benchmark helpers should be avoided if possible, as it enforces something to be implemented at runtime level, which is always a pain. If we can circumvent this with some design changes or additional methods in some traits, it should be done.

clangenb (commented):
Thanks for all the fixes! I only have a tiny remark and a question!

sage/pallet/src/lib.rs
DidacSF: I've managed to simplify the `BenchmarkHelper` but I still believe that removing it entirely would require changing the game transition related traits to a point in which they should not be. Let me know how do you like it now.

sage/pallet/src/mock.rs
```diff
@@ -357,11 +349,45 @@ impl AssetInspector for MockAssetMediator {
...
+		let asset_id_2 = Self::create_asset_for(account, season, seed * 2);
+		let asset_id_3 = Self::create_asset_for(account, season, seed * 3);
+		let asset_id_4 = Self::create_asset_for(account, season, seed * 4);
+		let asset_id_5 = Self::create_asset_for(account, season, seed * 5);
+
+		let asset_vec = vec![asset_id_1, asset_id_2, asset_id_3, asset_id_4, asset_id_5];
+
+		(ExampleTransitionId::BenchTransition, asset_vec)
+	}
+}
```
clangenb, line 386: I like this. Although, it might be misleading for the downstream implementors that we have a `BenchTransition` here. They will not have that, they simply must know the most expensive transition.

But this can be changed later.

clangenb (approved):
Looks good to me!

sage/pallet/src/lib.rs
clangenb: I think it looks good now. Yes, I agree that it will be hard to remove it entirely. I think it is good now as it is. 👍

PR #96, opened 2024-12-11
[AJUN-873] Introduce "Freemint" concept and call it "Voucher Points" to S.A.G.E.

DidacSF:
## Description

Adds the capability to use alternate payment methods "vouchers" in the FeeHandler implementations

## Type of changes

## Checklist


pallets/ajuna-awesome-avatars/benchmarking/src/mock.rs
```diff
@@ -53,7 +55,8 @@ frame_support::construct_runtime!(
 	pub struct Runtime {
 		System: frame_system = 0,
 		Balances: pallet_balances = 1,
-		Randomness: pallet_insecure_randomness_collective_flip = 2,
+		VoucherBalances: pallet_balances::<Instance1> = 2,
```
clangenb: I think this is a leftover.

pallets/ajuna-awesome-avatars/src/mock.rs
```diff
@@ -47,7 +49,8 @@ frame_support::construct_runtime!(
 	pub struct Test {
 		System: frame_system = 0,
 		Balances: pallet_balances = 1,
-		Randomness: pallet_insecure_randomness_collective_flip = 2,
+		VoucherBalances: pallet_balances::<Instance1> = 2,
```
clangenb: Also leftover

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -46,73 +47,104 @@ impl<AccountId, Balance> Payment<AccountId, Balance> {
 pub trait FeeHandler {
 	type AccountId;
 
-	type AssetId;
+	type Payment: Clone + Eq + Debug + TypeInfo + MaxEncodedLen + EncodeLike + Decode;
```
clangenb: I think payment is the wrong name here, understand that you wanted to distinguish it from the regular asset id though.

How about `PaymentType` or `PaymentKind`?

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -46,73 +47,104 @@ impl<AccountId, Balance> Payment<AccountId, Balance> {
...
-	// struct.
-	Assets: fungibles::Inspect<AccountId, Balance = W::Balance, AssetId = W::AssetId>
-		+ fungibles::Balanced<AccountId>,
-
-	W: WithdrawCredit<
+impl<AccountId, PaymentAssets, WPA, Affiliate, AffiliateMaxDistribution, Tournament> FeeHandler
+	for AssetGameFeeHandler<
+		AccountId,
+		PaymentAssets,
+		WPA,
```
clangenb: What is `WPA` supposed to stand for?

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -152,3 +155,37 @@ where
...
+		asset_id: Self::AssetId,
+		credit: Self::Balance,
+	) -> Result<Self::Credit, DispatchError> {
+		match asset_id {
+			WithdrawKind::Payment(payment_asset_id) =>
+				W::withdraw_credit(who, payment_asset_id, credit).map(Some),
+			WithdrawKind::Voucher => V::consume_vouchers_from(who, credit).map(|_| None),
+		}
+	}
+}
```
clangenb: Very nice. I like this design!

I would only suggest a small difference here. Instead, of returning an option here, I would directly return `Credit.zero`. In this case, we don't need to implement all those wrapper traits.

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -61,7 +62,7 @@ pub trait WithdrawCredit {
 	type AssetId: Clone + Eq + Debug + TypeInfo + MaxEncodedLen + EncodeLike + Decode;
 
 	type Assets;
-	type Balance: Balance;
+	type Balance: TokenBalance;
```
clangenb: Hmm, I think I prefer if this stays `Balance` because this is how it is consistently referred to in the polkadot-sdk.

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -1,12 +1,13 @@
+use crate::voucher_handler::VoucherHandler;
 use core::{fmt::Debug, marker::PhantomData};
 use frame_support::{
-	pallet_prelude::DispatchError,
+	pallet_prelude::{DispatchError, Encode},
 	traits::{
 		fungible,
-		fungible::Balanced,
+		fungible::Balanced as AssetBalanced,
```
clangenb: This is actually a wrong renaming.

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -72,21 +73,23 @@ pub trait WithdrawCredit {
 	) -> Result<Self::Credit, DispatchError>;
 }
 
-pub struct WithdrawNative<T>(PhantomData<T>);
+pub struct WithdrawNative<T, I>(PhantomData<(T, I)>);
 
-impl<T: pallet_balances::Config + frame_system::Config> WithdrawCredit for WithdrawNative<T> {
+impl<T: pallet_balances::Config<I> + frame_system::Config, I: 'static> WithdrawCredit
```
clangenb: Actually, we can simplify this by making `T: fungible::Balanced`, then we can omit the `I` generic type.

sage/pallet/src/lib.rs
```diff
@@ -621,7 +621,7 @@ pub mod pallet {
 		pub fn buy_asset(
 			origin: OriginFor<T>,
 			asset_id: AssetIdOf<T, I>,
-			payment_asset_id: Option<PaymentAssetIdOf<T, I>>,
+			payment: PaymentOf<T, I>,
 		) -> DispatchResult {
```
clangenb: It was a conscious choice to make this an option, as frontends can then decide to omit the payment type, so that they don't have to hassle with the enum if they don't want to. Do you see any drawback of that?

pallets/ajuna-tournament/src/impls/fee_handler.rs
```diff
@@ -1,19 +1,18 @@
...
+use ajuna_primitives::payment_handler::{DistributeFee, PaymentFee};
 use sp_arithmetic::traits::{CheckedDiv, Saturating};
-use sp_std::vec;
 
 impl<T: Config<I>, I: 'static> DistributeFee for Pallet<T, I> {
 	type AccountId = AccountIdFor<T>;
 	type Balance = BalanceOf<T, I>;
 	type FeeIdentifier = TournamentCategoryIdFor<T, I>;
-	type MaxDistributions = ConstU32<1>;
+	type FeeDistribution = PaymentFee<Self::AccountId, Self::Balance>;
```
clangenb, line 9: PaymentFee doesn't sound meaningful in English to me.  It would rather be `FeePayment`. However, because the struct itself is merely comprised of a beneficiary and an account, it think the generic name `Payment` is still fine. What was your reasoning behind changing the name?

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -190,20 +230,19 @@ where
...
-		account: &W::AccountId,
+		fee_credit: fungibles::Credit<AccountId, PaymentAssets>,
+		account: &AccountId,
 		identifier: &Tournament::FeeIdentifier,
-	) -> Result<W::Credit, DispatchError> {
+	) -> Result<fungibles::Credit<AccountId, PaymentAssets>, DispatchError> {
 		let mut final_fee = fee_credit;
 
-		if let Some(fee) = Tournament::distribute_fee(final_fee.peek(), account, identifier) {
-			assert_eq!(fee.len(), 1, "invalid fee provider implementation");
```
clangenb, line 200: I like that this is obsolete now.

pallets/ajuna-awesome-avatars/src/mock.rs
```diff
@@ -120,8 +154,14 @@ impl pallet_ajuna_awesome_avatars::Config for Test {
 	type FeeChainMaxLength = AffiliateMaxLevel;
 	type AffiliateHandler = Affiliates;
 	type TournamentHandler = Tournament;
-	type FeeHandler =
-		NativeGameFeeHandler<MockAccountId, Balances, WithdrawNative<Test>, Affiliates, Tournament>;
+	type FeeHandler = NativeGameFeeHandler<
+		MockAccountId,
+		Balances,
+		WithdrawCreditOrVoucher<WithdrawNative<Test, ()>, MockVoucherHandler>,
```
clangenb: Instead of using the mock here, can't we still just use the `WithdrawNative` here instead of adding the voucher?

pallets/ajuna-awesome-avatars/src/lib.rs
```diff
@@ -743,7 +744,7 @@ pub mod pallet {
 			if affiliate_config.mode == AffiliateMode::Open && affiliate_config.enabled_in_buy {
 				T::FeeHandler::withdraw_and_pay_fees(
 					&buyer,
-					(),
+					WithdrawKind::Payment(()),
```
clangenb: To preserve the current behavior for AAA now, I suggest keep using `()`, this should be possible isn't it?

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -220,56 +259,86 @@ where
...
+		Affiliate,
+		AffiliateMaxDistribution,
+		Tournament,
+	)>,
 }
 
-impl<AccountId, Balances, W, Affiliate, Tournament> FeeHandler
-	for NativeGameFeeHandler<AccountId, Balances, W, Affiliate, Tournament>
-where
+impl<AccountId, PaymentAsset, WPA, Affiliate, AffiliateMaxDistribution, Tournament> FeeHandler
```
clangenb: I think this should be `Balances` still.

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -220,56 +259,86 @@ where
...
 	>,
-
-	Affiliate: DistributeFee<AccountId = W::AccountId, Balance = W::Balance>,
-	Tournament:
-		DistributeFee<AccountId = AccountId, Balance = W::Balance, MaxDistributions = ConstU32<1>>,
 {
 	type AccountId = AccountId;
-	type AssetId = ();
-	type Balance = W::Balance;
+	type Payment = WPA::AssetId;
```
clangenb: I would like to add a comment here, that this can be `()` if no vouchers are intended.

primitives/payment-handler/src/lib.rs
clangenb, line 1: As mentioned in the other comment, I think by slightly changing `WithdrawCreditOrVoucher` we can delete this entire file.

primitives/payment-handler/src/mock.rs
```diff
@@ -176,13 +211,22 @@ impl EnsureWhitelistedAsset for WhitelistedAssets {
...
+		self
+	}
+}
 
 impl ExtBuilder {
 	pub fn build(self) -> sp_io::TestExternalities {
 		let config = RuntimeGenesisConfig {
 			system: Default::default(),
-			balances: pallet_balances::GenesisConfig { balances: vec![(ALICE, 100)] },
+			balances: pallet_balances::GenesisConfig::<Test, ()> { balances: vec![(ALICE, 100)] },
```
clangenb: I think that you can revert this.

sage/pallet/src/mock.rs
```diff
@@ -288,6 +303,24 @@ impl BenchmarkHelper<MockAccountId, MockSeasonId, AssetId, Asset, ExampleTransit
 
 		(ExampleTransitionId::BenchTransition, asset_vec)
 	}
+
+	fn create_payment() -> WithdrawKind<NativeOrWithId<u32>> {
```
clangenb: I would say that this should be `create_payment_type` or `create_payment_kind`

sage/pallet/src/mock.rs
```diff
@@ -65,8 +65,9 @@ frame_support::construct_runtime!(
 	pub struct Test {
 		System: frame_system = 0,
 		Balances: pallet_balances = 1,
-		PalletAssets: pallet_assets = 2,
-		Sage: pallet_sage = 3,
+		VoucherBalances: pallet_balances::<Instance1> = 2,
```
clangenb: I think we don't need this anymore with the current implementation

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -46,73 +47,104 @@ impl<AccountId, Balance> Payment<AccountId, Balance> {
...
-		payment_asset: Self::AssetId,
+		payment: Self::Payment,
 		base_fee: Self::Balance,
 		tournament_id: &Self::TournamentFeeIdentifier,
 		affiliate_id: &Self::AffiliateFeeIdentifier,
 		treasury_pot: &Self::AccountId,
 	) -> Result<(), DispatchError> {
 		// The credit may be in any asset as implemented by `WithdrawAsset`.
-		let fee_credit = W::withdraw_credit(payer, payment_asset, base_fee)?;
+		let fee_credit = WPA::withdraw_credit(payer, payment.clone(), base_fee)?.into_credit();
```
clangenb: There is now a very valid reason that this is 0. So we should do a check and do an early return in this case.

As mentioned in `withdraw_credit`, I don't think that we need the `into_credit` implementations if we internally convert it to `Credit.zero()`.

However, I just notice now. We could still return an option here, and then just early return if it is `None`. Either way, I don't think we need the `into_credit` boilerplate`.

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -220,56 +259,86 @@ where
...
+		WPA,
+		Affiliate,
+		AffiliateMaxDistribution,
+		Tournament,
+	> where
 	// This is satisfied by the `pallet-balances`.
-	Balances: fungible::Inspect<AccountId, Balance = W::Balance> + fungible::Balanced<AccountId>,
+	PaymentAsset:
+		fungible::Inspect<AccountId, Balance = WPA::Balance> + fungible::Balanced<AccountId>,
+	WPA: WithdrawCredit<AccountId = AccountId, Assets = PaymentAsset>,
```
clangenb: Tbh `WPA` confuses me more than `W`.

clangenb (commented):
Very good work! I did have quite some comments, but I do like the PR very much!

primitives/payment-handler/src/fee_handler.rs
DidacSF: I'll rename it to `PaymentKind`

primitives/payment-handler/src/fee_handler.rs
DidacSF: It means `WithdrawPaymentAssets`

pallets/ajuna-tournament/src/impls/fee_handler.rs
DidacSF: The name change came from the idea that it doesn't represent a number of distribution but more the concept of a fee distribution right now.

primitives/payment-handler/src/withdraw_credit.rs
DidacSF: Unfortunately , I wasn't able to keep it just as `T: fungible::Balanced<T::AccountId> + frame_system::Config>` for some reason the compiler says that `pallet_balances::Config<()>` doesn't implement `frame_system::Config` for now I just added the AccountId as an additional generic parameter.

primitives/payment-handler/src/fee_handler.rs
DidacSF: Changed to W

primitives/payment-handler/src/fee_handler.rs
DidacSF: Changed to W

sage/pallet/src/lib.rs
DidacSF: No, I probably changed it without thinking. I'll restore the previous code!

clangenb: Actually, I changed it after your rebase, so I guess it was just a rebase artifact. :D

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -46,118 +48,137 @@ impl<AccountId, Balance> Payment<AccountId, Balance> {
...
+		FeeDistribution = TournamentFeeDistribution<AccountId, W::Balance>,
+	>,
 {
 	/// Distributes an already withdrawn `fee_credit` to the affiliates of `account`.
 	///
 	/// Returns the remaining `fee_credit` after this operation.
 	fn try_propagate_chain_fee(
-		fee_credit: W::Credit,
-		account: &W::AccountId,
+		fee_credit: fungibles::Credit<AccountId, Assets>,
```
clangenb: Can we use `W::Credit` again? - it is easier to read. I think you have changed it because you had to in the first try.

There are a few other cases in this file, where this applies too.

sage/pallet/src/tests/extrinsics/buy_asset.rs
```diff
@@ -56,7 +56,11 @@ fn buy_should_work() {
 				asset_for_sale,
 				asset_price
 			));
-			assert_ok!(Sage::buy_asset(RuntimeOrigin::signed(ALICE), asset_for_sale, Some(NATIVE)));
+			assert_ok!(Sage::buy_asset(
+				RuntimeOrigin::signed(ALICE),
+				asset_for_sale,
+				MAYBE_NATIVE_PAYMENT
```
clangenb: Hmmm, I think `maybe` is misleading, as it implies that it could be none. I prefer to remove the prefix and then to rename the variable if we put none there.

sage/pallet/src/mock.rs
```diff
@@ -65,8 +65,8 @@ frame_support::construct_runtime!(
 	pub struct Test {
 		System: frame_system = 0,
 		Balances: pallet_balances = 1,
-		PalletAssets: pallet_assets = 2,
-		Sage: pallet_sage = 3,
+		PalletAssets: pallet_assets = 3,
+		Sage: pallet_sage = 4,
```
clangenb: Nitpick: we skip an index now.

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -69,86 +67,97 @@ pub trait WithdrawCredit {
...
-	type Credit = fungibles::Credit<Self::AccountId, pallet_assets::Pallet<T>>;
+impl<AccountId, Fungibles> WithdrawCredit for WithdrawFungibles<AccountId, Fungibles>
+where
+	Fungibles: fungibles::Balanced<AccountId>,
+{
+	type AccountId = AccountId;
+	type AssetId = Fungibles::AssetId;
+	type Assets = Fungibles;
+	type Balance = Fungibles::Balance;
+	type Credit = fungibles::Credit<Self::AccountId, Fungibles>;
```
clangenb, line 112: Very nice, maximally generified!

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -69,86 +67,97 @@ pub trait WithdrawCredit {
...
 }
 
-pub struct WithdrawNative<T>(PhantomData<T>);
+pub struct WithdrawNative<AccountId, T>(PhantomData<(AccountId, T)>);
 
-impl<T: pallet_balances::Config + frame_system::Config> WithdrawCredit for WithdrawNative<T> {
-	type AccountId = T::AccountId;
+impl<AccountId, T> WithdrawCredit for WithdrawNative<AccountId, T>
+where
+	T: fungible::Balanced<AccountId>,
```
clangenb: Can we be consistent with the rest of the file and rename `T` -> `Fungible`?

primitives/payment-handler/src/tests.rs
```diff
@@ -179,17 +181,58 @@ mod asset_fee_handler {
...
+				// check that no fee has also been allocated to the affiliates
+				assert_eq!(Assets::balance(WHITELISTED_ASSET_ID, BOB), 0);
+				assert_eq!(Assets::balance(WHITELISTED_ASSET_ID, CHARLIE), 0);
+				assert_eq!(Assets::balance(WHITELISTED_ASSET_ID, DAVE), 0);
+
+				// check that the requested vouchers have been deducted from storage
+				let alice_current_vouchers = VOUCHERS
+					.with_borrow(|voucher_store| voucher_store.get(&ALICE).copied())
+					.expect("Should contain remaining vouchers");
+				assert_eq!(alice_current_vouchers, alice_initial_vouchers - fee);
```
clangenb, line 221: I think that the implementation should be: 1 Voucher == 1 Execution. The production implementation has no conversion from the fee to the vouchers, as there exists no market. So my proposition is the only way it makes sense to me.

primitives/payment-handler/src/tests.rs
```diff
@@ -256,17 +333,47 @@ mod asset_fee_handler {
...
+				assert_eq!(Assets::balance(WHITELISTED_ASSET_ID, ALICE), alice_balance_before);
+				assert_eq!(Assets::balance(WHITELISTED_ASSET_ID, fee_beneficiary), 0);
+
+				// check that the requested vouchers have been deducted from storage
+				let alice_current_vouchers = VOUCHERS
+					.with_borrow(|voucher_store| voucher_store.get(&ALICE).copied())
+					.expect("Should contain remaining vouchers");
+				assert_eq!(alice_current_vouchers, alice_initial_vouchers - fee);
+			});
+		}
```
clangenb: This is actually very interesting. I guess it makes sense the way it is implemented, i.e, successfully deposit, but don't deposit anything. However, instead of `_works`, I would call this test: 

`withdraw_and_deposit_into_treasury_with_vouchers_does_not_err`

And we should add a comment that this is just the defensive behavior, as we don't want an error.

primitives/payment-handler/src/tests.rs
```diff
@@ -504,17 +720,47 @@ mod native_fee_handler {
 			});
 		}
 
+		#[test]
+		fn withdraw_and_deposit_into_treasury_works_with_vouchers() {
```
clangenb: Same comment as for above test

clangenb (commented):
Nice, I like the implementation a lot! There are only a few nitpicks left. :)

clangenb (approved):
Nice thanks for all the fixes! Looks good to me now!

PR #97, opened 2024-12-18
[AJUN-910] Update benchmarks to V2 syntax

DidacSF:
## Description

Updates all benchmarking code in the repo to use the `frame_benchmarking::v2` syntax

## Type of changes

## Checklist

### DIFF
--- a/pallets/ajuna-nft-staking/benchmarking/src/lib.rs
+++ b/pallets/ajuna-nft-staking/benchmarking/src/lib.rs
@@ -19,7 +19,7 @@
 
 mod mock;
 
-use frame_benchmarking::benchmarks;
+use frame_benchmarking::v2::*;
 use frame_support::{
 	pallet_prelude::*,
 	traits::{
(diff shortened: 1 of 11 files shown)


sage/pallet/src/benchmarking.rs
```diff
@@ -211,101 +227,103 @@ benchmarks! {
...
-			RawOrigin::Signed(acc_1.clone()).into(),
-			asset_id.clone())
-		.expect("Should lock asset");
-	}: _(RawOrigin::Signed(acc_1), asset_id.clone())
-	verify {
-		assert_last_event::<T, ()>(Event::AssetUnlocked {
-			asset_id,
-			lock: expected_lock,
-		})
+		Pallet::<T, ()>::lock_asset(RawOrigin::Signed(acc_1.clone()).into(), asset_id.clone())
```
clangenb: We have to replace `()` with `I` otherwise it will fail if the runtime tries to use the non-default instance.

pallets/ajuna-nft-staking/benchmarking/src/lib.rs
```diff
@@ -101,19 +101,20 @@ fn assert_last_event<T: Config>(avatars_event: Event<T>) {
...
-			.try_for_each(|item_id| mint_item::<T>(&creator, REWARD_COLLECTION, item_id))?;
+			.try_for_each(|item_id| mint_item::<T>(&creator, REWARD_COLLECTION, item_id))
+			.expect("Account should be created");
 	}
 	Creator::<T>::put(&creator);
 	Ok(creator)
 }
 
 fn create_contract_collection<T: Config>(creator: &T::AccountId) -> DispatchResult {
-	create_collection::<T>(creator)?;
```
clangenb: This seems to me like it was a search and replace error. No reason to replac the `?` here.

Btw. We could have chosen to return a `Result<(), BenchmarkError>` in the benchmarks, then we can keep the `?`. I have no strong opinion here though.

clangenb (commented):
Some tiny remarks. Otherwise, it looks great! Finally, we did that step. :)

clangenb (approved):
Nice, looks good to me!

PR #98, opened 2024-12-23
[AJUN-939] Move nft fee hanlding outside pallet-nft-transfer

DidacSF:
## Description

Moves the `handle_prepare_asset_fee` logic outside the `pallet-nft-transfer` and integrated into the `FeeHandler` implementations.

## Type of changes

## Checklist


primitives/payment-handler/src/fee_handler.rs
```diff
@@ -322,8 +413,47 @@ where
...
+		from: &Self::AccountId,
+		fees_recipient: &Self::AccountId,
+	) -> Result<(), DispatchError> {
+		if let Some(fee) = NftFee::compute_fee(from, asset) {
+			W::Assets::resolve(fees_recipient, fee)
+				.map_err(|_| DispatchError::Token(TokenError::FundsUnavailable))?;
+		}
+
+		Ok(())
+	}
```
clangenb: I am not really happy how this looks like. I tried to improve it a bit, but using the full features of our fee handler will make the nft-transfer-pallet more complex, every attempt to simplify that fails at the `PaymentType` the latest.

So, to simplify our design, I would recommend doing the following:
1. For now, we decide that the nft-transfer fee must be native to simplify things.
2. instead of the fee handler, we use the fungible trait (replacement of the deprecated currency trait)
3. We make e simple transfer of the fee to the service account, and the fee could be just be a field of the nft-transfer's general config storage.

I think it is ok to say that only our core gameing features can be paid with other currencies besides native. I would argue this caters for 99% of all game related payments, and therefore I would not complicate things needlessly.

Do you agree @DidacSF?

What do you think?

DidacSF: Hmm, I do agree that trying to integrate the nft payment into the FeeHandler makes a it a tad too complicated. So yes, for now its probably best if we do what you propose. I'll get to it!

pallets/ajuna-nft-transfer/src/lib.rs
```diff
@@ -93,6 +104,8 @@ pub mod pallet {
 
 		type AccountManager: AccountManager<AccountId = AccountIdFor<Self>>;
 
+		type FeeHandler: InspectFungible<AccountIdFor<Self>> + MutateFungible<AccountIdFor<Self>>;
```
clangenb: Hmm, I kind of think it would be more intuitive to just call this associated type `Fungible`, what do you think?

clangenb (commented):
I only have one remark. Otherwise, it looks good to me!

pallets/ajuna-nft-transfer/src/lib.rs
DidacSF: Hmm yeah, I agree!

PR #99, opened 2025-01-09
HeroJam implementation as the SAGE pallet example

DidacSF:
## Description

Implements the code in [this repository](https://github.com/ajuna-network/Ajuna.SAGE/tree/master/Ajuna.SAGE.Game.HeroJam) to be used as the SAGE pallet example game implementation

## Type of changes

## Checklist


sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,294 @@
...
+			2 => WorkType::Gather,
+			3 => WorkType::Build,
+			4 => WorkType::Travel,
+			5 => WorkType::Train,
+			6 => WorkType::Fight,
+			7 => WorkType::Rest,
+			_ => WorkType::Hunt,
+		}
+	}
+}
```
clangenb, line 73: The question is if it is maybe better to have a `try_from` instead of just defaulting to some variant.

sage/pallet/src/mock.rs
```diff
@@ -204,7 +215,7 @@ pub struct MockAssetMediator;
 impl AssetManager for MockAssetMediator {
```
clangenb, line 181: I think it is important that the mock uses the sage-pallet here. We need to do this how we would do it in the actual runtime, so that we don't run into any unexpected type issues.

sage/example-transition/src/rules/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, MaxEncodedLen, TypeInfo)]
+pub enum Rule<AccountId, TransitionRuleset, AssetHandler> {
+	Not(Box<Self>),
+	Or(Box<Self>, Box<Self>),
+	And(Box<Self>, Box<Self>),
+	AssetCount(u8),
+	IsOwnerOf(AccountId, PhantomData<AssetHandler>),
+	Transition(TransitionRuleset),
+}
```
clangenb: I wonder if this enum is actually beneficial. I mean, we are maybe trying to reduce friction for non-rust devs, but we are talking to devs at least. I could imagine that it is actually less complicated if we just offer functions instead of introducing our enum here.

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,294 @@
...
+			HeroAction::Sleep(_) | HeroAction::Work(_, _) => vec![
+				HeroJamRuleset::AssetCount(1),
+				HeroJamRuleset::IsOwnerOf(account_id.clone(), PhantomData),
+				HeroJamRuleset::Transition(HeroJamRule::AllAssetType(AssetType::Hero)),
+				HeroJamRuleset::Transition(HeroJamRule::CanStateChange(PhantomData)),
+			],
+			_ => vec![],
+		};
+
+		Self::verify(account_id, asset_ids, ruleset.as_slice())
```
clangenb: Hmm, I mean it looks very nice, but is not the most efficient. I wonder if we should replace it with something like:

```rust
		match transition_id {
			HeroAction::Create => {
			      check_asset_count(assets, 1)?;
			      assert_has_asset_type(hero)?;
                              Ok(());
			},
			HeroAction::Sleep(_) | HeroAction::Work(_, _) => {
			    check_asset_count(assets, 1)?;
			    check_ownership(account_id, assets)?;
			    check_all_asset_type(AssetType::Hero, assets)?;
			    can_change_state(asset)?;
			    Ok(())			
			},
		}
```

The benefits of this approach would be:
* For simple rules, we avoid generics
* We can nicely separate general rules, and rules specific to hero jam
* Non-rust devs don't need to hassle with rust concepts like enums
* More efficient (no iterating through a ruleset, although the overhead is negligible for small vecs like this)

What is your take here?

sage/example-transition/src/transition/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+			AccountId = AccountId,
+			AssetId = asset::AssetId,
+			Asset = asset::Asset<BlockNumber>,
+		> + AssetInspector<
+			AccountId = AccountId,
+			AssetId = asset::AssetId,
+			Asset = asset::Asset<BlockNumber>,
+		>,
+	ChainHandler: ChainInspector<BlockNumber = BlockNumber>,
+{
```
clangenb, line 40: I do wonder if it is easier for a dev if we just supply an aggregated associated type here. I guess, we could offer both, so someone who uses all the sage features can just use the aggregated trait bound, and someone who uses only a part of it, could for instance only use the `AssetHandler`.

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,294 @@
...
+						asset.state_type = StateType::Sleep;
+						asset.state_sub_type = SleepType::Normal as u8;
+						asset.state_sub_value = *sleep_time as u8;
+						asset.state_change_block_number =
+							Self::get_block_time_for_action(sleep_time);
+
+						Ok(vec![TransitionOutput::Mutated(asset_ids[0], asset::Asset::from(asset))])
+					},
+				}
+			},
```
clangenb: We could use a method here:

```
impl Asset {
   fn try_hero_jam(self) -> Option<HeroJamAsset>

   fn as_hero_jam(&self) -> Option<&HeroJamAsset>
}
```

Then we don't have to use the match statement everywhere.

Edit: Actually, we could even use the `TryInto` trait for that.

sage/example-transition/src/asset/hero_jam.rs
```diff
@@ -0,0 +1,46 @@
...
+	pub balance: u32,
+}
+
+impl<BlockNumber> GetId<AssetId> for HeroJamAsset<BlockNumber> {
+	fn get_id(&self) -> AssetId {
+		// TODO: Improve
+		// TODO: Maybe we should rethink the whole 'GetId' trait?
+		0
+	}
+}
```
clangenb: Hmm, yes. I agree. We should think about what exactly we want the id for and when.

clangenb (commented):
Looks very good. I don't really have to say anything regarding the design, I think we are already at the details!

sage/api/src/rules.rs
```diff
@@ -1,7 +1,24 @@
...
+	owner: &AccountId,
+) -> Result<(), RuleError>
+where
+	Manager: AssetManager<AccountId = AccountId, AssetId = AssetId>,
+{
+	assets
+		.iter()
+		.all(|asset_id| Manager::ensure_ownership(owner, asset_id).is_ok())
+		.then_some(())
+		.ok_or(RuleError::AssetOwnership)
```
clangenb: I was wondering here if it might be better understandable if we use something like this:

```rust
	if assets.iter().all(|asset_id| Manager::ensure_ownership(owner, asset_id).is_ok()) {
		Ok(())
	} else {
		Err(RuleError::AssetOwnership)
	}
```
What do you think?

sage/example-transition/src/asset/hero_jam.rs
```diff
@@ -39,8 +40,16 @@ pub struct HeroJamAsset<BlockNumber> {
...
+
+impl<BlockNumber> TryFrom<Asset<BlockNumber>> for HeroJamAsset<BlockNumber> {
+	type Error = ();
+
+	fn try_from(value: Asset<BlockNumber>) -> Result<Self, Self::Error> {
+		match value.asset_variant {
+			AssetVariant::HeroJam(hero_jam_asset) => Ok(hero_jam_asset),
+		}
 	}
```
clangenb: We can actually implement the `From` too here. Or am I mistaken?

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -240,55 +221,53 @@ impl<AccountId, BlockNumber, AssetHandler, ChainHandler> SageGameTransition
...
-			],
-			_ => vec![],
-		};
-
-		Self::verify(account_id, asset_ids, ruleset.as_slice())
+	) -> Result<(), RuleError> {
+		match transition_id {
+			HeroAction::Create => {
+				ensure_asset_length(asset_ids, 0)?;
+				if ensure_account_has_asset_of_type::<_, _, AssetHandler>(
```
clangenb: I would add a function that does the opposite too. This would simplify the game devs code greatly, i.e.,:

```rust
ensure_account_has_not_asset_of_type(...)
```

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -1,91 +1,119 @@
...
+{
+	if let Ok(asset) = Inspector::get_asset(asset_id) {
+		if let Ok(hero_jam_asset) = HeroJamAsset::try_from(asset) {
+			Ok(hero_jam_asset)
+		} else {
+			Err(RuleError::Other { error: ASSET_NOT_HERO_JAM })
+		}
+	} else {
+		Err(RuleError::Other { error: ASSET_NOT_FOUND })
+	}
```
clangenb: I prefer to avoid double nesting, and do something like this, and I think we can in general make this more elegant:

```rust
let asset = Inspector::get_asset(asset_id).ok_or(RuleError::Other { error: ASSET_NOT_FOUND })?;
asset.try_into().map_err(|_| RuleError::Other { error: ASSET_NOT_HERO_JAM })
```

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -1,91 +1,119 @@
...
+{
+	let hero_jam_assets = try_cast_to_hero_jam_assets::<_, _, Inspector>(assets)?;
+
+	hero_jam_assets
+		.iter()
+		.all(|hero_jam_asset| hero_jam_asset.asset_type == asset_type)
+		.then_some(())
+		.ok_or(RuleError::Other { error: ASSETS_NOT_ALL_SAME_TYPE })
+}
+
```
clangenb: This is inefficient, we iterate through the assets, convert them to hero jam assets, and collect (=O(n)!) and then we iterate through them again? ;) I think we can do better.

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -1,91 +1,119 @@
...
-pub enum Rule<AccountId, AssetHandler, ChainHandler> {
-	AllAssetType(hero_jam::AssetType),
-	AllStateType(hero_jam::StateType),
-	CanStateChange(PhantomData<ChainHandler>),
-	AccountHasAssetOfType(hero_jam::AssetType, PhantomData<(AccountId, AssetHandler)>),
+	hero_jam_assets
+		.iter()
+		.all(|hero_jam_asset| hero_jam_asset.state_type == state_type)
+		.then_some(())
+		.ok_or(RuleError::Other { error: ASSETS_NOT_ALL_SAME_STATE })
```
clangenb: Same inefficiency as above

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -1,91 +1,119 @@
...
+pub const ASSET_NOT_FOUND: u8 = 200;
+pub const ASSET_NOT_HERO_JAM: u8 = 201;
+
+pub const ASSETS_NOT_ALL_SAME_TYPE: u8 = 100;
+pub const ASSETS_NOT_ALL_SAME_STATE: u8 = 101;
+pub const ASSET_CANNOT_STATE_CHANGE: u8 = 102;
+pub const ASSET_HERO_NOT_IN_ACCOUNT: u8 = 103;
+pub const ASSET_HERO_ALREADY_IN_ACCOUNT: u8 = 104;
+
+fn try_cast_to_hero_jam_asset<AccountId, BlockNumber, Inspector>(
```
clangenb: `cast` is a wrong term here. It is used with the `as` keyword. Into/From functions are called conversions.

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -1,91 +1,119 @@
...
+		.filter(|asset_id| {
+			if let Ok(asset) = try_cast_to_hero_jam_asset::<_, _, Inspector>(asset_id) {
+				asset.asset_type == asset_type
+			} else {
+				false
+			}
+		})
+		.count()
+		.eq(&1)
+	{
```
clangenb: I am pretty sure that we can replace this whole thing with a `.first(predicate)` call.

sage/example-transition/src/rules/hero_jam.rs
clangenb: Actually, this is the wrong term anyhow, we just get the asset as hero jam. I would rename this into `try_get_hero_jam()`

sage/api/src/error.rs
```diff
@@ -16,14 +21,23 @@ impl AsErrorCode for u8 {
...
+impl AsErrorCode for RuleError {
 	fn as_error_code(&self) -> u8 {
 		match self {
-			Error::InvalidTransitionId => 0,
-			Error::InvalidAssetLength => 1,
-			Error::TransferError => 2,
-			Error::FeeError => 3,
-			Error::Transition { error } => *error,
+			RuleError::AssetLength => 0,
+			RuleError::AssetOwnership => 1,
```
clangenb: Just to be extra unambiguous. I would recommend that the TransitionError and the RuleError don't have overlapping error types.

clangenb (commented):
I have some low-level nitpicks, but otherwise it looks very good!

sage/example-transition/src/asset/hero_jam.rs
DidacSF: We can yes, but in the future if we have multiple variants then would need to implement `TryFrom` instead, so I decided to just implement `TryFrom` to remark this fact.

clangenb: Yes, but we can always do an exhaustive match statement here, don't we?

DidacSF: Thing is `From` is infallible, and right now it would not fail, but once `asset_variant` contains more than just hero jam then we will need to remove that `From` implementation. That's why I don't think there's much point in implementing `From`

clangenb: Ahhh, sorry. I read that the conversion is in the other direction: HeroJam -> Asset. You are perfectly correct.

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,266 @@
...
+					account_id,
+					AssetType::Hero,
+				)
+			},
+			HeroAction::Sleep(_) | HeroAction::Work(_, _) => {
+				ensure_asset_length(asset_ids, 1)?;
+				ensure_owner_of::<_, _, AssetHandler>(asset_ids, account_id)?;
+				ensure_all_asset_type::<_, _, AssetHandler>(asset_ids, AssetType::Hero)?;
+				ensure_can_state_change::<_, _, AssetHandler, ChainHandler>(&asset_ids[0])
+			},
```
clangenb: I just realized, in all these functions, we read the asset from storage redundantly. I guess, we want to read all the assets once at the beginning of the rule verification and then just run checks on them without reading again from storage.

clangenb: Actually, it is still redundant, because we read them in the transition execution too. Shall we switch to a paradigm like?

```rust
fn execute_transition(...) {
    let assets = get_assets()?;

    // maybe we have to pass the asset_ids too for certain checks
    let Self::verify_rule(&assets, ...)?;

    Self::do_transition(assets, ...)
}
```

What do you think? Tomorrow me optimization?

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -0,0 +1,134 @@
...
+) -> bool
+where
+	Inspector: AssetInspector<AccountId = AccountId, AssetId = AssetId, Asset = Asset<BlockNumber>>,
+{
+	Inspector::iter_assets_from(account_id).any(|asset_id| {
+		if let Ok(hero_jam) = try_get_hero_jam::<_, _, Inspector>(&asset_id) {
+			hero_jam.asset_type == asset_type
+		} else {
+			false
+		}
```
clangenb: Here the functional variant might be better readable:

```rust
try_get_hero_jam::<_, _, Inspector>(&asset_id)
   .any(|asset_id| 
          try_get_hero_jam::<_, _, Inspector>(&asset_id)
             .map(|hero_jam| hero_jam.asset_type == asset_type)
             .unwrap_or(false)
   )
```

Not sure, do whatever you feel is better.

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,266 @@
...
+
+	fn get_resource_energy(state_type: StateType, state_sub_type: u8, action_time: u8) -> i32 {
+		match state_type {
+			StateType::None => 0,
+			StateType::Sleep => match SleepType::from(state_sub_type) {
+				SleepType::Normal => -(action_time as i32),
+				_ => 0,
+			},
+			StateType::Work => match WorkType::from(state_sub_type) {
+				WorkType::Hunt => -((action_time as i32) * 6),
```
clangenb, line 259: We have to use the saturated conversions here to prevent any panics. Also, this makes me realize, we have to implement our own a panic catcher, we can allow that another game's code makes our chain panic. I will add this to our list.

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	fn get_resource_fatigue(state_type: StateType, state_sub_type: u8, action_time: u8) -> i32 {
+		match state_type {
+			StateType::None => 0,
+			StateType::Sleep => match SleepType::from(state_sub_type) {
+				SleepType::Normal => -((action_time as i32) * (action_time as i32) * 11),
+				_ => 0,
+			},
+			StateType::Work => (action_time as i32) * 8,
```
clangenb, line 247: Never use signed types if not necessary. You can omit the type annotations on the left hand side if you put  `* 8u32*

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -0,0 +1,266 @@
...
+			},
+			HeroAction::Sleep(sleep_time) => {
+				match AssetHandler::get_asset(&asset_ids[0])
+					.map_err(|_| TransitionError::Transition { error: 0 })?
+					.asset_variant
+				{
+					AssetVariant::HeroJam(mut asset) => {
+						asset.state_type = StateType::Sleep;
+						asset.state_sub_type = SleepType::Normal as u8;
+						asset.state_sub_value = *sleep_time as u8;
```
clangenb: Why do we have these type casts?

clangenb (commented):
I had another round of review. Still looks very good, just some details!

sage/example-transition/src/lib.rs
```diff
@@ -1,7 +1,19 @@
...
+	pub use crate::{
+		asset::{
+			hero_jam::{AssetSubType, AssetType, HeroJamAsset, StateType},
+			Asset, AssetId, AssetVariant,
+		},
+		benchmarks::GameBenchmarkHelper,
+		filter::GameFilter,
+		transition::{GameTransition, TransitionIdentifier},
+	};
+}
```
clangenb, line 19: Good practice 👍

sage/example-transition/src/rules/hero_jam.rs
```diff
@@ -5,13 +5,14 @@ use crate::asset::{
...
-pub const ASSETS_NOT_ALL_SAME_TYPE: u8 = 100;
-pub const ASSETS_NOT_ALL_SAME_STATE: u8 = 101;
-pub const ASSET_CANNOT_STATE_CHANGE: u8 = 102;
-pub const ASSET_HERO_NOT_IN_ACCOUNT: u8 = 103;
-pub const ASSET_HERO_ALREADY_IN_ACCOUNT: u8 = 104;
+pub(crate) const ASSETS_NOT_ALL_SAME_TYPE: u8 = 100;
+pub(crate) const ASSETS_NOT_ALL_SAME_STATE: u8 = 101;
+pub(crate) const ASSET_CANNOT_STATE_CHANGE: u8 = 102;
+pub(crate) const ASSET_HERO_NOT_IN_ACCOUNT: u8 = 103;
+pub(crate) const ASSET_HERO_ALREADY_IN_ACCOUNT: u8 = 104;
```
clangenb: In general, I would probably make most things 100% public in such a library.

sage/pallet/src/tests/extrinsics/state_transition.rs
```diff
@@ -115,7 +121,10 @@ fn state_transition_should_reject_rule_verification_failure() {
 			// sage/example-transition/src/generic.rs
 			// requires the input assets to be of length 1
```
clangenb: I guess the comment is outdated now.

clangenb (commented):
Only some tiny details, but otherwise it looks good!

PR #101, opened 2025-01-20
Add the option to deposit funds into assets

clangenb:
This allows the option to allocate any kind of fungible asset to one of our sage assets.

However, currently the way we use it it is only going to be the native asset. The user can define any fungible asset that should be consumed, but the native asset will be allocated to the sage asset. This is configurable in the runtime however.


sage/pallet/src/trait_impls/asset_manager.rs
```diff
@@ -105,3 +111,83 @@ impl<T: Config<I>, I: 'static> AssetInspector for Pallet<T, I> {
...
+
+impl<T: Config<I>, I: 'static> AssetFundsManager for Pallet<T, I> {
+	type AccountId = AccountIdOf<T>;
+	type AssetId = AssetIdOf<T, I>;
+	type Balance = BalanceOf<T, I>;
+
+	fn inspect_asset_funds(asset_id: &Self::AssetId) -> Self::Balance {
+		let asset_hash = <T as frame_system::Config>::Hashing::hash_of(&asset_id);
+		let asset_account =
+			Self::AccountId::decode(&mut asset_hash.encode().as_bytes_ref()).unwrap();
```
clangenb: This is a quick implementation. I was wondering if we should introduce a trait like:

* `get_account_id` for the asset? But I do like that we don't need the asset here, and just the asset id.

However, if we want some convenient linking between the asset and it's account id, I guess that we have to do such a trait.

DidacSF: Hm, I would prefer to avoid constraining the Assets as much as possible, so I believe this is fine.

Additionally, have you though of, instead of deriving an account from the asset hash, to simply lock the funds for the owner of the asset when "depositing" funds onto it? That way we don't even need an extra account to manage the asset's funds like you have now.

DidacSF (approved):
In general I like the approach! What I'm missing right now is the integration in the example, but this can be done later.

sage/pallet/src/trait_impls/asset_manager.rs
clangenb: First, I was against this. But with the current code I have right now, it makes a lot of sense, and it would actually even remove the need of an account id derivation in the first place! L

Very good idea!

primitives/payment-handler/src/tests.rs
```diff
@@ -405,13 +403,11 @@ mod asset_fee_handler {
 					.expect("Should contain remaining vouchers");
 
 				assert_noop!(
-					TestAssetFeeHandler::withdraw_and_pay_fees(
+					TestAssetFeeHandler::withdraw_and_deposit_into(
```
clangenb, line 406: This was an error. We used the wrong method here for the test.

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -69,7 +69,7 @@ pub trait FeeHandler {
 
 	/// Withdraws the `amount` denominated in `payment` and allocates it fully to the
 	/// `treasury_pot`.
-	fn withdraw_and_deposit_into_treasury(
+	fn withdraw_and_deposit_into(
```
clangenb, line 72: I noticed that this functio is currently more generic than the treasury. So I thought I will rename it as such. However, we have to see in the future if it actually makes sense to have it here.

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -123,30 +123,30 @@ where
...
 				Self::try_propagate_tournament_fee(fee_credit, payer, tournament_id)?;
 
 			let remaining_credit2 =
 				Self::try_propagate_chain_fee(remaining_credit, payer, affiliate_id)?;
 
-			Self::deposit_into_treasury(treasury_pot, remaining_credit2)
+			Self::deposit(treasury_pot, remaining_credit2)
 		} else {
+			// This is only none if the fee was paid with a voucher.
+			// In this case we simply do nothing.
```
clangenb, line 138: I was having issues remembering why this is ok XD

primitives/payment-handler/src/fee_handler.rs
```diff
@@ -238,14 +238,40 @@ where
 
 		Ok(final_fee)
 	}
+}
+
+impl<AccountId, Assets, W, Affiliate, MaxAffiliates, Tournament>
+	AssetGameFeeHandler<AccountId, Assets, W, Affiliate, MaxAffiliates, Tournament>
```
clangenb, line 244: Just extracted from above. Initially, I did that because I thought I can re-use the fee handler for general transferring, but I decided then that this is not a good idea. I kept the minor refactor though.

primitives/payment-handler/src/transfer_fungible.rs
```diff
@@ -0,0 +1,297 @@
+use crate::{IdentifyVoucherOrAssetId, WithdrawCredit};
```
clangenb, line 1: This is where the meat is.

primitives/payment-handler/src/withdraw_credit.rs
```diff
@@ -114,26 +113,104 @@ where
...
 		)
 		.map(Some)
 	}
 }
 
-#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone, MaxEncodedLen, TypeInfo, Default)]
+#[derive(Debug, Encode, Decode, PartialEq, Eq, Clone, MaxEncodedLen, TypeInfo)]
 pub enum WithdrawKind<AssetId> {
 	Payment(AssetId),
-	#[default]
```
clangenb, line 133: Default was bad in my opinion here, as this might be unexpected, and I actually think it should be the native currency. Instead of using and requiring default, and I opted to introduce another trait `NativeId`.

DidacSF (approved):

PR #105, opened 2025-01-31
Add Genesis Config to pallet-sage and -seasons

clangenb:


sage/pallet/src/lib.rs
```diff
@@ -103,6 +103,41 @@ pub mod pallet {
...
+					PlayerSeasonConfigs::<T, I>::insert(
+						organizer,
+						season_id,
+						PlayerConfig {
+							inventory_tier: InventoryTier::One,
+							locks: Locks::all_unlocked(),
+						},
+					);
+				}
+			}
```
clangenb, line 137: I have some doubts about implicitly assuming values here instead of explicitly setting values, but I guess you did this to avoid hassling with more trait bounds?

DidacSF (approved):

PR #107, opened 2025-01-31
[CI] more explicit use of nightly version in fmt job

clangenb:
This updates the nightly, which has slightly different fmt settings. Hence, some files had to be formatted.

I have always been bothered that I got different results on my machine, but this should be resolved now.

Side-effect:
* The fmt job takes now 17secs instead of 50 secs 🥳

### DIFF
--- a/.github/workflows/check-pull-request.yml
+++ b/.github/workflows/check-pull-request.yml
@@ -24,18 +24,24 @@ env:
   CARGO_ARGS: ${{ github.ref == 'refs/heads/develop' && '--release' || '' }}
   CARGO_TERM_COLOR: always
   SKIP_WASM_BUILD: 1
+  RUST_NIGHTLY_VERSION: 2024-11-19
 
 jobs:
   fmt:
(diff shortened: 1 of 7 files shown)


.github/workflows/check-pull-request.yml
```diff
@@ -24,20 +24,26 @@ env:
...
+      - name: Install nightly toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          targets: "wasm32-unknown-unknown"
+          components: "rustfmt"
+          toolchain: "nightly-${{env.RUST_NIGHTLY_VERSION}}"
+
         # some settings are only available in nightly.
-      - run: cargo +nightly fmt --all -- --check
+      - run: cargo +nightly-$RUST_NIGHTLY_VERSION fmt --all -- --check
```
clangenb, line 46: Now it is very explicit, which nightly is used here.

DidacSF (approved):

PR #108, opened 2025-01-31
Integrate adding funds to hero jam asset

clangenb:
This PR also extracts the `TransitionConfig` as its standalone storage type. I think this is better in a separate of concerns, and it is best practice to separate storage values wherever they are not inherently tied together.

I think that we should separate custom game types from our types in general, so that we can decouple our changes from their changes better.

There are a few todos that emerged during this PR, but I consider this out of scope and I have created JIRA issues for that.

### DIFF
--- a/sage/example-transition/src/transition/hero_jam.rs
+++ b/sage/example-transition/src/transition/hero_jam.rs
@@ -220,9 +220,10 @@ where
 				);
 
 				if work_type == &WorkType::Hunt {
+					let hunting_reward = Sage::get_transition_config().hunting_reward;
+
 					// Also do the accounting on the asset, but this is not necessary per se.
-					asset.balance = asset.balance.saturating_add(10);
-					let transition_config = Sage::get_transition_config();
+					asset.balance = asset.balance.saturating_add(hunting_reward);
 
 					// Todo: how to handle dispatch errors in transitions
 					Sage::deposit_funds_to_asset(
@@ -231,7 +232,7 @@ where
 						// Todo: how to pass the FungibleAssetId that was used as payment for this
 						<Sage::FungiblesAssetId as NativeId>::get_native_id(),
 						// Todo: use balance type in transition config
-						transition_config.hunting_reward,
+						hunting_reward,
 					)
 					.expect("transferring to asset failed");
 				}
--- a/sage/example-transition/src/transition/hero_jam.rs
+++ b/sage/example-transition/src/transition/hero_jam.rs
@@ -14,7 +14,9 @@ use sage_api::{
 };
 
 use crate::transition::GameTransitionConfig;
-use ajuna_primitives::sage_api::SageApi;
+use ajuna_primitives::{
+	asset_manager::AssetFundsManager, payment_handler::NativeId, sage_api::SageApi,
+};
 use core::marker::PhantomData;
 use frame_support::{
 	pallet_prelude::{Decode, Encode, MaxEncodedLen, TypeInfo},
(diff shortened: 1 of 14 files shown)


sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -97,18 +101,21 @@ impl From<u8> for SleepType {
...
+	HeroJamTransition<AccountId, BlockNumber, AssetHandler, ChainHandler, Sage>
 where
 	AccountId: Member + Codec,
 	BlockNumber: BlockNumberT,
 	AssetHandler: AssetManager<AccountId = AccountId, AssetId = AssetId, Asset = Asset<BlockNumber>>
 		+ AssetInspector<AccountId = AccountId, AssetId = AssetId, Asset = Asset<BlockNumber>>,
 	ChainHandler: ChainInspector<BlockNumber = BlockNumber>,
+	Sage: SageApi<TransitionConfig = GameTransitionConfig>
+		+ AssetFundsManager<AccountId = AccountId, AssetId = AssetId, Balance = u64>,
+	<Sage as AssetFundsManager>::FungiblesAssetId: NativeId,
```
clangenb, line 118: I think I would like to go back to the initial design that I laid out:

1. Define one aggregated `SageApi` trait which implements all these traits here

Benefits:
1. No redundant associated type constraints, i.e., define `AccountId=AccountId` like 5 times
2. Less error prone for the game dev
3. Only generic parameter, it is anyhow 3 times the `SageFacade` that we pass as the generic type (see tests).

What do you think?

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -213,7 +220,20 @@ where
...
+					// Also do the accounting on the asset, but this is not necessary per se.
 					asset.balance = asset.balance.saturating_add(10);
+					let transition_config = Sage::get_transition_config();
+
+					// Todo: how to handle dispatch errors in transitions
+					Sage::deposit_funds_to_asset(
+						&asset_id,
+						&account_id,
+						// Todo: how to pass the FungibleAssetId that was used as payment for this
+						<Sage::FungiblesAssetId as NativeId>::get_native_id(),
```
clangenb, line 233: I see no other option than passing an extra argument to the state transition. Do you have another idea?

sage/pallet/src/tests/extrinsics/state_transition.rs
```diff
@@ -5,18 +5,19 @@
...
 // the Free Software Foundation, either version 3 of the License, or
 // (at your option) any later version.
-
+//
 // This program is distributed in the hope that it will be useful,
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU Affero General Public License for more details.
-
+//
```
clangenb, line 13: The IDE always tried to auto import stuff here. XD

sage/example-transition/src/transition/hero_jam.rs
```diff
@@ -213,7 +220,20 @@ where
 				);
 
 				if work_type == &WorkType::Hunt {
+					// Also do the accounting on the asset, but this is not necessary per se.
 					asset.balance = asset.balance.saturating_add(10);
+					let transition_config = Sage::get_transition_config();
+
+					// Todo: how to handle dispatch errors in transitions
```
clangenb, line 228: Created JIRA issues for those todos

DidacSF (approved):

PR #112, opened 2025-02-04
Add aggregated `SageApi` and introduce a macro for implementing it in the runtime

clangenb:


sage/example-transition/src/asset/mod.rs
```diff
@@ -7,22 +7,24 @@ pub mod hero_jam;
...
-	HeroJam(hero_jam::HeroJamAsset<BlockNumber>),
+pub enum AssetVariant<BlockNumber, Balance> {
+	HeroJam(hero_jam::HeroJamAsset<BlockNumber, Balance>),
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Encode, Decode, MaxEncodedLen, TypeInfo)]
-pub struct Asset<BlockNumber> {
-	pub asset_variant: AssetVariant<BlockNumber>,
+pub struct Asset<BlockNumber, Balance> {
+	pub asset_variant: AssetVariant<BlockNumber, Balance>,
```
clangenb, line 16: I think the generics are a hassle. To simplify the lives of downstream implementors, I think we should expose all the types of our runtime in ajuna primitives, so that they can use the actual concrete type in the implementation. The only downside to that is that the tests need to use the same types.

DidacSF: I think exposing the types in ajuna-primitives may be a good idea. But for the example I would still use the generics.

DidacSF (approved):
I like it a lot! The macro is also very nice!

PR #113, opened 2025-02-11
Adjust transition arguments: make extra optional

clangenb:


sage/api/src/error.rs
```diff
@@ -1,7 +1,7 @@
 pub enum TransitionError {
-	InvalidTransitionId,
```
clangenb: I do not think that this error makes sense, rust's type system makes it impossible to mess this up.

sage/api/src/traits.rs
```diff
@@ -19,17 +19,28 @@ pub trait SageGameTransition {
...
+	///
+	/// If you don't need fungible tokens in your transition you can set this as ().
+	type PaymentFungible;
+
 	fn do_transition(
 		transition_id: &Self::TransitionId,
 		account_id: &Self::AccountId,
 		assets_ids: &[Self::AssetId],
-		extra: &Self::Extra,
+		extra: Option<Self::Extra>,
```
DidacSF, line 33: I don't think forcing the `extra` argument to be an `Option` is a good idea, since you can already define that in your implementation if needed. 

Also, if you don't need the extra at all you can just define it as the empty type and be done with it.

DidacSF (changes requested):

sage/api/src/traits.rs
clangenb: I can see your point, but what if only one of your transitions needs an extra, and all others don't need it? I see that your point is that the extra type itself can be an option, but I do think that this is a bit more obscure and harder to understand and the game implementor might think that this is something mandatory.

I will think a bit more about it to see if there are actual hard arguments for either option.

DidacSF: yeah, the thing is that fundamentally the extra argument can be part of the transition_id entirely anyways, so maybe the best approach is to simplify the trait by removing it

clangenb: I was thinking of the design you mentioned before, but then I think it is wrong to separate the transition and the id. There should be no more ID then, just a rich rust enum:

```rust
pub enum HeroTransition {
    CreateHero, // no args needed
    Work { asset: ..., other_args }
    ...
}

impl HeroTransition {
    fn execute(self) -> Result<TransitionResult, TransitionError> {
        match self {
           Create => // create
           Work { .... } => // do work      
        }
    }
}
```

This approach would already implicitly check 50% of the rules we have, and make them obsolete, and the only downside is the relatively bad rust enum support in other programming languages.

clangenb: hmmm, I keep coming back to this idea, and I start to think that bad enum support in other languages is not enough reason to not do it in rust.

PR #116, opened 2025-02-18
[AJUN-1014] CasinoJam as the SAGE example

DidacSF:
## Description

Implements the logic from https://github.com/ajuna-network/Ajuna.SAGE/tree/master/Ajuna.SAGE.Game.CasinoJam as a SAGE game.

## Type of changes

## Checklist


sage/example-transition/src/asset/hero_jam.rs
clangenb, line 1: Is it the consensus to abort hero jam?

sage/pallet/src/tests/extrinsics/state_transition.rs
```diff
@@ -193,16 +152,17 @@ fn hero_hunt_transition_add_funds_to_balance_works() {
...
+			let added_balance = TokenType::T10.get_value_for(MultiplierType::V1).into();
 			assert_eq!(
 				Balances::free_balance(ALICE),
-				initial_balance - transition_fee - hunting_reward
+				initial_balance - transition_fee - added_balance
 			);
-			assert_eq!(Balances::free_balance(Sage::assets_funds_pot()), hunting_reward);
+			// Since in the mock the treasury account and the assets fund pot account
+			// are the same, they share the balance deposited to them.
+			assert_eq!(Balances::free_balance(Sage::assets_funds_pot()), added_balance + 2);
```
clangenb, line 162: Instead of writing a comment here, shouldn't we better separate those pots even in the pot?

sage/example-transition/src/asset.rs
```diff
@@ -0,0 +1,286 @@
...
+impl<BlockNumber> Asset<BlockNumber>
+where
+	BlockNumber: BlockNumberT,
+{
+	pub fn new_player(player_id: AssetId, genesis: BlockNumber) -> Self {
+		Self {
+			id: player_id,
+			collection_id: ASSET_COLLECTION_ID,
+			genesis,
+			variant: AssetVariant::Player(PlayerVariant::Human(HumanVariant::default())),
```
clangenb: Are these type names copied from c#? I think that except for `AssetVariant` it is better to remove the `-variant` suffix, as I believe that it is actually misleading information. The `HumanVariant` or the `TrackerVariant` for instance are actually independent, full-fledged structs. It is a coincidence that they are also the fields of an enum variant.

sage/example-transition/src/asset.rs
```diff
@@ -0,0 +1,286 @@
...
+		}
+	}
+
+	pub fn try_as_seat(&mut self) -> Result<&mut SeatVariant<BlockNumber>, TransitionError> {
+		match &mut self.variant {
+			AssetVariant::Seat(seat_variant) => Ok(seat_variant),
+			_ => Err(TransitionError::Transition { code: ASSET_VARIANT_IS_NOT_SEAT }),
+		}
+	}
+}
```
clangenb, line 117: I think these are good helpers, and I like the error code solution. 👍

sage/example-transition/src/asset.rs
```diff
@@ -0,0 +1,286 @@
...
+				MachineSubVariant::Bandit(_) => VariantType::Machine(MachineType::Bandit),
+			},
+			AssetVariant::Seat(_) => VariantType::Seat,
+		}
+	}
+
+	pub fn is_variant(&self, variant_type: VariantType) -> bool {
+		self.get_variant_type() == variant_type
+	}
+}
```
clangenb, line 156: This is a bit confusing to me. Do I understand it correctly that the `VariantType` and its fields are only here so we can match an asset in:

* tests
* filters
?

If so (actually also if not), we need to add some documentation here.

sage/example-transition/src/asset.rs
```diff
@@ -0,0 +1,286 @@
...
+pub enum MachineSubVariant {
+	Bandit(BanditVariant),
+}
+
+impl Default for MachineSubVariant {
+	fn default() -> Self {
+		Self::Bandit(BanditVariant::default())
+	}
+}
+
```
clangenb: I think we have to be very careful about defaults for enums. Do they really make sense? I prefer to not have a default implementation if it is not future-proof, or not really needed.

sage/example-transition/src/rules.rs
```diff
@@ -0,0 +1,92 @@
+use crate::{
+	asset::{Asset, AssetId, VariantType},
+	error::*,
+};
+
+use ajuna_primitives::sage_api::SageApi;
+use sage_api::TransitionError;
+
+#[allow(dead_code)]
```
clangenb: Dead code is not linted, and no warnings are emitted. I prefer to make it pub in this case, so that it is better checked.

sage/example-transition/src/rules.rs
```diff
@@ -0,0 +1,92 @@
...
+	if let Some((_, asset)) = assets.get(asset_index) {
+		if asset.variant.is_variant(variant_type) {
+			Ok(())
+		} else {
+			Err(TransitionError::Transition { code: ASSET_TYPE_NOT_VALID })
+		}
+	} else {
+		Err(TransitionError::Transition { code: ASSET_NOT_FOUND })
+	}
+}
```
clangenb: I would prefer this version here, as it is more readable

```rust
let (_, asset) = assets.get(asset_index).ok_or_else(|| ..error)?;

if asset.variant.is_variant(variant_type) {
	Ok(())
} else {
	Err(TransitionError::Transition { code: ASSET_TYPE_NOT_VALID })
}

```

sage/example-transition/src/rules.rs
```diff
@@ -0,0 +1,92 @@
...
+	filter_fn: F,
+) -> bool
+where
+	F: Fn((AssetId, Asset<BlockNumber>)) -> bool,
+	Sage: SageApi<AccountId = AccountId, AssetId = AssetId, Asset = Asset<BlockNumber>>,
+{
+	Sage::iter_assets_from(account_id).all(|asset| !filter_fn(asset))
+}
+
+#[allow(dead_code)]
```
clangenb: ditto

sage/example-transition/src/rules.rs
```diff
@@ -0,0 +1,92 @@
...
+fn account_has_no_asset<AccountId, BlockNumber, Sage, F>(
+	account_id: &AccountId,
+	filter_fn: F,
+) -> bool
+where
+	F: Fn((AssetId, Asset<BlockNumber>)) -> bool,
+	Sage: SageApi<AccountId = AccountId, AssetId = AssetId, Asset = Asset<BlockNumber>>,
+{
+	Sage::iter_assets_from(account_id).all(|asset| !filter_fn(asset))
+}
```
clangenb: I think we can remove the inline here, I am not sure if it really benefits us, same with the other inline.

sage/example-transition/src/transition/enums.rs
```diff
@@ -0,0 +1,63 @@
...
+		10_u32.pow(*self as u32)
+	}
+
+	pub fn get_value_for(&self, amount: MultiplierType) -> u32 {
+		self.as_value().saturating_mul(amount as u32)
+	}
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default, Encode, Decode, MaxEncodedLen, TypeInfo)]
+pub enum MultiplierType {
```
clangenb, line 37: maybe add a `#[repr(u8)]` here to ensure that we never have a variant panicking when we cast it to `u16`?

sage/pallet/src/mock.rs
```diff
@@ -66,7 +68,8 @@ frame_support::construct_runtime!(
 		System: frame_system = 0,
 		Balances: pallet_balances = 1,
 		PalletAssets: pallet_assets = 2,
-		Sage: pallet_sage = 3,
+		HashRandomness: pallet_insecure_randomness_collective_flip = 3,
```
clangenb: I kind of feel that plain `Randomness` is better 🙈

sage/example-transition/src/transition/mod.rs
```diff
@@ -1,73 +1,611 @@
...
+		TransitionConfig = CasinoJamTransitionConfig,
+		HashOutput = H256,
+	>,
+{
+	fn try_get_asset(asset_id: &AssetId) -> Result<Asset<BlockNumber>, TransitionError> {
+		let asset = Sage::get_asset(asset_id)
+			.map_err(|_| TransitionError::Transition { code: ASSET_NOT_FOUND })?;
+
+		Ok(asset)
+	}
```
clangenb, line 80: If we only have these functions because of the error type, we should change sage to already return such an error.

In this case, I think we can just add an actual error variant `AssetNotFound`.

sage/example-transition/src/transition/mod.rs
```diff
@@ -1,73 +1,611 @@
...
+		let hash_base = Sage::random_hash(b"create_asset");
+		let current_block_number = Sage::get_current_block_number();
+
+		hash_base
+			.0
+			.into_iter()
+			.fold(0 as AssetId, |acc, value| acc.saturating_add(value as AssetId))
+			.saturating_add(current_block_number.saturated_into())
+			.saturating_add(nonce as AssetId)
+	}
```
clangenb, line 135: I think that this is not good practice: performing operations on a low-influence hash function output to generate another random value removes the low-influence property. We should only allow two options here:

* If the asset-it is a byte array, we should directly take the hash
* If it is a monotonic counter

Maybe both options can be configured via a pallet SAGE associated type, and then we can simply call `next_asset_id()` from the transition, which might either be a random, or an increased value.

sage/example-transition/src/transition/mod.rs
```diff
@@ -1,73 +1,611 @@
...
+			},
+			CasinoAction::Kick => {
+				let assets = Self::try_get_assets(asset_ids)?;
+
+				ensure_asset_length(asset_ids, 3)?;
+				ensure_owner_of::<_, _, Sage>(&asset_ids[0..1], account_id)?;
+				ensure_asset_type_at(&assets, VariantType::Player(PlayerType::Human), 0)?;
+				ensure_asset_type_at(&assets, VariantType::Player(PlayerType::Human), 1)?;
+				ensure_asset_type_at(&assets, VariantType::Seat, 2)?;
+
```
clangenb, line 218: It hurts me to see that an actual transition enum, would make 100% of these checks unnecessary.

sage/example-transition/src/transition/mod.rs
clangenb: I think it would be awesome if we can slightly change the sage interface, so that we can just remove all of the above boilerplate code. I think this is possible without too much effort. (but this can be follow-up work.)

sage/pallet/src/tests/example/create.rs
```diff
@@ -0,0 +1,125 @@
...
+					assert!(asset_2.variant.is_variant(VariantType::Player(PlayerType::Tracker)));
+
+					let player_1 = asset_1.try_as_player().expect("Should be player");
+					let human = player_1.try_as_human().expect("Should be human");
+					let player_2 = asset_2.try_as_player().expect("Should be player");
+					let tracker = player_2.try_as_tracker().expect("Should be tracker");
+
+					(human, tracker)
+				} else {
+					assert!(asset_1.variant.is_variant(VariantType::Player(PlayerType::Tracker)));
```
clangenb, line 59: I am having a hard time to follow. Is this test not deterministic?

clangenb (commented):
Wow, all in all I think it is very great work. I like seeing that many tests, and I think that the code is in general very readable! I do see a few issues because (as you said) you had to cater for C# inheritance, but I actually do think that the compromises are bearable.

There are only some minor low-level nitpicks, and a few details about the `SageApi`. The latter I would actually prefer to fix in follow-up PRs, to make this a `CasinoJam` PR only.

Cool PR.

sage/pallet/src/tests/extrinsics/state_transition.rs
DidacSF: This only happens if the AccountId is a small type like in the Mock, so for production code this is not an issue, separating both accounts is more work than is worth I believe right now.

sage/example-transition/src/asset/hero_jam.rs
DidacSF: Theoretically, yes.

sage/example-transition/src/asset.rs
DidacSF: Well I use them to represent the C# inheritance structure, so naming them variant is a way of saying that their existence is based around the fact that they are only ever used as part of an enum

sage/example-transition/src/asset.rs
DidacSF: I think the code is self-descriptive enough.

sage/example-transition/src/transition/enums.rs
DidacSF: I don't think that will have any influence on that, since the variants are already marked with the proper value.

sage/example-transition/src/transition/mod.rs
DidacSF: I agree that maybe we should move the asset id generation to Sage itself

sage/example-transition/src/transition/mod.rs
DidacSF: Well,  even with the transition enum some of these checks would still need to be performed, but thats life

sage/pallet/src/tests/example/create.rs
DidacSF: It should be, I just added the check to cover all bases

sage/example-transition/src/transition/mod.rs
DidacSF: The reason for the error is that the error `get_asset` returns is `DispatchError` whereas we expect a `TransitionError` in the Sage transition trait, so we have three options here:
* We define a error enum for the SageApi trait instead of `DispatchError`
* We change the expected error type of the sage transition trait to `DispatchError`
* We leave things as they are and make error conversions if necessary.

sage/example-transition/src/transition/enums.rs
clangenb: hmmm, I mean this would be just in case that someone tries to add a number > 2^15, but yeah. It is a nitpick.

sage/example-transition/src/asset.rs
clangenb: Hmm, ok maybe. I do agree that documentation should only be done when strictly necessary.

sage/pallet/src/tests/extrinsics/state_transition.rs
clangenb: Yeah, I do wonder anyhow about the benefits of using something else than AccountId32 for the tests tbh. XD

clangenb (approved):
As discussed, we can do the other things later (if at all).

Looks good to me!

PR #118, opened 2025-03-07
[sage] add AssetId provider and ensure unique asset ids for CasinoJam

clangenb:
This should be deployed quite quickly, actually. Currently, there is a chance that we create duplicate asset ids.


sage/example-transition/src/transition/mod.rs
```diff
@@ -122,16 +122,9 @@ where
...
-
-		hash_base
-			.0
-			.into_iter()
-			.fold(0 as AssetId, |acc, value| acc.saturating_add(value as AssetId))
-			.saturating_add(current_block_number.saturated_into())
-			.saturating_add(nonce as AssetId)
+	fn generate_asset_id() -> Result<AssetId, TransitionError> {
+		Sage::create_next_asset_id()
+			.ok_or_else(|| TransitionError::CouldNotCreateAssetId)
```
clangenb: I think we should improve our error conversion. There is too much boilerplate code in the SageTransition, and I think we can actually easily avoid that. I created a JIRA ticket for this.

DidacSF (approved):
I like the idea!

One thing that I will ask is if you can also apply it to the sage-playground repo?

PR #120, opened 2025-04-22
Add boilerplate code for testing; introduce sage-testing crate

clangenb:
The idea is that a game dev can very easily have access to a full `SageApi` by simply calling a few lines of boilerplate code. The changes should be very self-explanatory, and I have integrated the usage already in the pallet-sage (as much as possible).

Remarks:
* It offers the capability of injecting a randomness provider that makes the outcome predictable for tests.

Follow ups (JIRA issues exists):
* pallet-ajuna-seasons: improve genesis config handling
* TestExternalities: Write a generic function that sets a sensible set of defaults for testruntimes


sage/pallet/src/mock.rs
clangenb, line 1: Most changes here come from using the types from sage-testing. The biggest change is that we start to use the actual account id type. I think this is better anyhow to get tests which match the real scenario as close as possible.

sage/pallet/src/mock.rs
```diff
@@ -182,57 +161,26 @@ impl SeasonManager for MockSeasonManager {
...
-impl_runtime_sage_api!(
+impl_test_runtime_sage_api!(
 	TestSageEngine,
 	Test,
 	DefaultSageInstance,
 	MockSeasonManager,
 	AssetId,
 	MockAsset,
 	CasinoJamTransitionConfig,
+	Randomness,
```
clangenb, line 175: Here, we could inject the `TestRandomness` from sage-testing, however, we have a few tests here that ensure that the randomness is indeed random, so I chose to keep the actual randomness. The test randomness can be used when testing the game transitions.

sage/testing/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+//!   TestRuntime,
+//!   Sage,
+//!   AjunaSeasons,
+//!   GameAssetId,
+//!   GameAsset,
+//!   TransitionConfig,
+//!   Randomness,
+//!   H256
+//! );
+//! ```
```
clangenb, line 63: This piece of code, and defining the TestExternalities should be the only thing needed to get a full test setup.

I am thinking about how to get a generic `TestExternalities` setup, but I think this should actually be very straight forward with generics.

DidacSF (approved):

